PR #2, opened 2018-03-14
Don't assume availability of get_nprocs_conf()

jserv:
get_nprocs_conf() is Linux-specific, and on macOS, it is absent.
Our own implementation of get_nprocs_conf() on macOS is provided
for compatibility.

### DIFF
--- a/src/pow_sse.c
+++ b/src/pow_sse.c
@@ -29,10 +29,27 @@
 #include <stdio.h>
 #include <string.h>
 #include <pthread.h>
-#include <sys/sysinfo.h>
 #include "pow_sse.h"
 #include "./hash/curl.h"
 #include "constants.h"
+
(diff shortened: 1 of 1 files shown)


src/pow_sse.c
```diff
@@ -29,11 +29,29 @@
...
 #include "constants.h"
 
+/* Required for get_nprocs_conf() on Linux */
+#ifdef NPROCS
+#include <sys/sysinfo.h>
+#endif
+
+/* On Mac OS X, define our own get_nprocs_conf() */
+#if defined (__APPLE__) || defined(__FreeBSD__)
+#include <pthread.h>
```
furuame: Duplicate inclusion of <pthread.h>

jserv: DONE in force-updated commit.

PR #16, opened 2018-03-16
Eliminate inconsistent pointers in signedness

jserv:
TODO: explicitly use uint8_t and int8_t in <stdint.h> rather than
platform-dependant char.


furuame:
I have fixed it by explicitly using int8_t.

PR #17, opened 2018-03-18
Source tree re-organization

jserv:
For consistency consideration, the following changes are done:
  * Directory `test` is renamed to `tests`.
  * No more directory `src/trinary` and `src/hash` exist.
  * All test cases have canonical prefix `test-` to distinguish.


jserv:
Regarding `$(PYFLAGS)`, the misbehavior was caused by the negligence of this commit.

@chenwei-tw, you can generate customized `tests/test-multi_pow` according to GPU availability.

PR #18, opened 2018-03-18
Better Java integration

jserv:
1. Move Java build rules out of top-level Makefile;
2. Use javah to generate JNI header from latest Java source;
3. Defer to JAVA_HOME setting;


furuame:
This feature is great! Thanks a lot.

jserv:
Alternatively, developers should be able to specify IRI source path rather than downloading from GitHub.

furuame:
[verified] Compatible with IRI.

PR #48, opened 2018-06-07
Fix function validateTrytes

sean-lin-tw:
A valid tryte should be composed of uppercase letters and number '9'


furuame:
Thanks :)

PR #62, opened 2018-08-28
Fix issue #61

### LINKED ISSUE #61: Unable to pass multi-pow test item with FPGA accelerator
Error messages with FPGA accelerator:

```
root@lampa:~/dcurl# make BUILD_FPGA_ACCEL=1 check                               
make: Warning: File 'mk/python.mk' has modification time 1534568560 s in the fue
  CC    build/curl.o                                                            
  CC    build/constants.o                                                       
  CC    build/trinary.o                                                         
  CC    build/dcurl.o                                                           
  CC    build/implcontext.o                                                     
  CC    build/pow_c.o                                                           
  CC    build/pow_fpga_accel.o                                                  
  CC    build/test-trinary.o                                                    
  LD    build/test-trinary                                                      
*** Validating build/test-trinary ***                                           
        [ Verified ]                                                            
  CC    build/test-curl.o                                                       
  LD    build/test-curl                                                         
*** Validating build/test-curl ***                                              
        [ Verified ]                                                            
  CC    build/test-dcurl.o                                                      
  LD    build/test-dcurl                                                        
*** Validating build/test-dcurl ***                                             
[  409.187196] Get interrupt                                                    
        [ Verified ]                                                            
  LD    build/libdcurl.so                                                       
*** Validating build/test-multi-pow ***                                         
[  413.727193] Get interrupt                                                    
[  413.745242] Get interrupt                                                    
[  413.992183] Get interrupt                                                    
[  414.190852] Get interrupt                                                    
[  414.291223] Get interrupt                                                    
[  414.386504] Get interrupt                                                    
[  414.517357] Get interrupt                                                    
Unhandled exception in thread started by <function call_dcurl at 0xb5a75300>    
Traceback (most recent call last):                                              
  File "./build/test-multi-pow", line 47, in call_dcurl                         
    trytes = TryteString(ret[:2673])                                            
TypeError: 'NoneType' object is not subscriptable 
```
and then type ctrl+c

```
Traceback (most recent call last):                                            
  File "./build/test-multi-pow", line 80, in <module>                           
    testing()                                                                   
  File "./build/test-multi-pow", line 71, in testing                            
    while join_list[i].locked(): pass                                           
KeyboardInterrupt                                                               
mk/common.mk:27: recipe for target 'build/test-multi-pow.done' failed           
make: *** [build/test-multi-pow.done] Interrupt                                 
make: *** Deleting intermediate file 'build/test-trinary.o'                     
make: *** Deleting intermediate file 'build/test-curl.o'                        
make: *** Deleting intermediate file 'build/test-dcurl.o'  
```

ajblane:
This issue is caused by the automatically performed adjustment
of the file offset by using low-level I/O. It leads the file
offset to be not correct when reentering the PoWFPGAAccel function
to execute low-level I/O. Therefore, it needs to reset file offset.
Close #61

### DIFF
--- a/src/pow_fpga_accel.c
+++ b/src/pow_fpga_accel.c
@@ -54,11 +54,11 @@ static bool PoWFPGAAccel(void *pow_ctx)
     Trits_t *object_trits = trits_from_trytes(object_trytes);
     if (!object_trits)
         return false;
-    
-    lseek(ctx->in_fd,0,0);
-    lseek(ctx->ctrl_fd,0,0);
-    lseek(ctx->out_fd,0,0);
-    
+
+    lseek(ctx->in_fd, 0, 0);
+    lseek(ctx->ctrl_fd, 0, 0);
+    lseek(ctx->out_fd, 0, 0);
+
     if (write(ctx->in_fd, (char *) object_trits->data, transactionTrinarySize) <
         0)
         return false;
--- a/src/pow_fpga_accel.c
+++ b/src/pow_fpga_accel.c
@@ -54,7 +54,11 @@ static bool PoWFPGAAccel(void *pow_ctx)
     Trits_t *object_trits = trits_from_trytes(object_trytes);
     if (!object_trits)
         return false;
-
+    
+    lseek(ctx->in_fd,0,0);
+    lseek(ctx->ctrl_fd,0,0);
+    lseek(ctx->out_fd,0,0);
+    
     if (write(ctx->in_fd, (char *) object_trits->data, transactionTrinarySize) <
         0)
         return false;


jserv (changes requested):
You have to explain why this change was intended to fix and how it was wrong in Git commit message.

jserv (changes requested):
If you would like to perform such cleanup, send another pull request to make reviewers concentrate on all functioned changes.

jserv (commented):
Can you use `cppcheck` or some static analyzers to validate as well?

src/pow_fpga_accel.c
```diff
@@ -41,7 +41,11 @@ static bool PoWFPGAAccel(void *pow_ctx)
     Trits_t *object_trits = trits_from_trytes(object_trytes);
     if (!object_trits)
         return false;
-
+    
+    lseek(ctx->in_fd,0,0);
```
furuame: Please maintain the coding style.

src/pow_fpga_accel.c
```diff
@@ -41,7 +41,11 @@ static bool PoWFPGAAccel(void *pow_ctx)
...
+    lseek(ctx->ctrl_fd,0,0);
+    lseek(ctx->out_fd,0,0);
```
furuame: So what the reason results in this issue ?

ajblane: This is because "The adjustment of the file offset and the write operation are performed as an atomic step" [0]. Therefore, the file offset is not correct when reentering  PoWFPGAAccel function.

[0] http://man7.org/linux/man-pages/man2/write.2.html

furuame: I don't think "write operation are performed as atomic step" is related to this issue, while in your implementation only one thread could enter the PoWFPGAAccel function.

furuame:
@ajblane Please only include the commit `dccead2` in this Pull Request as @jserv said. (also fix the coding style mentioned above)

src/pow_fpga_accel.c
furuame: I understand the problem is that the file offset is modified. Hence, it need to be initialized before doing FPGA Pow. Please explain the problem and solution in this Pull Request Description.

PR #65, opened 2018-09-02
Sanity check for OpenCL Library

furuame:


mk/opencl.mk
```diff
@@ -1,7 +1,13 @@
-# FIXME: perform sanity checks before compilation
-OPENCL_LIB ?= /usr/local/cuda-9.1/lib64
+OPENCL_LIB_PATH ?= /usr/local/cuda/lib64
+
+OPENCL_LIB := $(shell test -d $(OPENCL_LIB_PATH); echo $$?)
```
jserv: Can you validate some critical headers as well?

mk/opencl.mk
```diff
@@ -1,7 +1,13 @@
...
+    $(error "Please specify the path to your OpenCL library on your platform")
+endif
 
 CFLAGS += \
         -DENABLE_OPENCL \
-        -I$(OPENCL_PATH)/include
-LDFLAGS += -L$(OPENCL_LIB) -lOpenCL
+        -I$(OPENCL_LIB_PATH)/../include
+
+LDFLAGS += -L$(OPENCL_LIB_PATH) -lOpenCL
```
jserv: Before assigning `$(LDFLAGS)`, we shall ensure the presence of OpenCL runtime library in advance.

furuame: Do you mean check the presence of `libOpenCL.so` ?

mk/opencl.mk
furuame: Which headers do you refer? Something like `CL/cl.h` in this case?

mk/opencl.mk
jserv: Yes, exactly.

mk/opencl.mk
jserv: Users might specify other OpenCL implementation rather than the one shipped by NVIDIA SDK.  Take [Beignet](https://www.freedesktop.org/wiki/Software/Beignet/) for example, it is an implementation dedicated to Intel Graphics and suitable for CI purpose. However, it is hard for current scripts to detect its availability.

furuame: I would give it a try.

jserv:
@marktwtn, Be aware of the availbility of [Intel OpenCL driver](https://github.com/KhronosGroup/SyclParallelSTL/issues/69) which might be required by Travis-CI.

jserv:
Related: #69

mk/opencl.mk
furuame: done.

mk/opencl.mk
furuame: The availability of Beignet should be included in another issue #69.

furuame:
@jserv Should the OpenCL sanity check include the Travis-CI coverage?

jserv:
> @jserv Should the OpenCL sanity check include the Travis-CI coverage?

CI changes could be addressed in another pull request. Let's concentrate on OpenCL validation here.

Issue #47, opened 2018-05-27
Show the hash rate of each implementation


furuame:
Show the hash rate of each implementation by using `BUILD_BENCH` option

furuame:
@marktwtn Please reference the implementation in [iotaledger/giota](https://github.com/iotaledger/giota)

marktwtn:
The current implementation of getting the hash times would be calling an API
`getHashTimes(ImplContext *impl_ctx, void *pow_ctx)`

It would call the corresponding hardware-related function.
Take **CPU - SSE** as an example:
```
static uint64_t PoWSSE_getHashTimes(void *pow_ctx)
{
    uint64_t count = 0;
    PoW_SSE_Context *ctx = (PoW_SSE_Context *) pow_ctx;
    Pwork_struct *pitem = ctx->pitem;

    for (int i = 0; i < ctx->num_threads; i++) {
        if (pitem[i].ret >= 0 )
            count += (uint64_t)pitem[i].ret;
        else
            count += (uint64_t)(-(pitem[i].ret) + 1);
    }

    return count;
}
```

The **GPU** implementation is still under investigation.
The **FPGA** implementation would be completed by @ajblane .

furuame:
@marktwtn About hash times of the OpenCL, you can take a look from giota. It seems trivial.

https://github.com/iotaledger/giota/blob/e34a409290248e6ff2205b06062d702690910f04/pow_cl.go#L94

marktwtn:
`hash rate = hash count / PoW execution time`

There are two ways to get the `PoW execution time`.

1. Subtract the time before and after the `doThePoW` function.
2. Create an API like `getHashCount` to get the `PoW execution time`.
The time would be slightly accurate since the time subtraction would be done in the `doThePoW` function.
It excludes the function call, the PoW initialization part and other things. It focuses on the PoW execution time.

However, if the time unit is second, I think the `PoW execution time` may have no difference at all.

I prefer the method 2.
Is there any suggestion?

And I am thinking that the `hash rate` should be got with the API, too.

furuame:
Method 2 might be better to get the time.

But I'm not sure do we really need "three" APIs to achieve the goal of this enhancement?

marktwtn:
> But I'm not sure do we really need "three" APIs to achieve the goal of this issue?

How about

1. Create the `getHashRate` API and remove the others.

or

[...]

PR #72, opened 2018-09-20
Implement the getHashTimes() API

marktwtn:
Obtain the hash times mentioned in #47 with API.

The FPGA part will be implemented by @ajblane .

### DIFF
--- a/src/clcontext.h
+++ b/src/clcontext.h
@@ -41,6 +41,7 @@ typedef struct {
     cl_ulong max_memory;
     size_t num_work_group;
     KernelInfo kernel_info;
+    uint64_t hashTimes;
 } CLContext;
 
 enum {
--- a/src/implcontext.c
+++ b/src/implcontext.c
@@ -64,3 +64,8 @@ int8_t *getPoWResult(ImplContext *impl_ctx, void *pow_ctx)
 {
     return impl_ctx->getPoWResult(pow_ctx);
 }
+
+uint64_t getHashTimes(ImplContext *impl_ctx, void *pow_ctx)
+{
+    return impl_ctx->getHashTimes(pow_ctx);
+}
--- a/src/pow_avx.c
+++ b/src/pow_avx.c
@@ -681,6 +681,22 @@ static int8_t *PoWAVX_getPoWResult(void *pow_ctx)
     return ret;
 }
 
+static uint64_t PoWAVX_getHashTimes(void *pow_ctx)
+{
+    uint64_t count = 0;
+    PoW_AVX_Context *ctx = (PoW_AVX_Context *) pow_ctx;
+    Pwork_struct *pitem = ctx->pitem;
+
+    for (int i = 0; i < ctx->num_threads; i++) {
+        if (pitem[i].ret >= 0 )
+            count += (uint64_t)pitem[i].ret;
+        else
+            count += (uint64_t)(-(pitem[i].ret) + 1);
+    }
+
+    return count;
+}
+
 ImplContext PoWAVX_Context = {
     .context = NULL,
     .description = "CPU (Intel AVX)",
@@ -693,4 +709,5 @@ ImplContext PoWAVX_Context = {
     .freePoWContext = PoWAVX_freePoWContext,
     .doThePoW = PowAVX,
     .getPoWResult = PoWAVX_getPoWResult,
+    .getHashTimes = PoWAVX_getHashTimes,
 };
--- a/src/pow_sse.c
+++ b/src/pow_sse.c
@@ -457,6 +457,22 @@ static int8_t *PoWSSE_getPoWResult(void *pow_ctx)
     return ret;
 }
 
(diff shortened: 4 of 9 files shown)


src/clcontext.h
```diff
@@ -41,6 +41,7 @@ typedef struct {
     cl_ulong max_memory;
     size_t num_work_group;
     KernelInfo kernel_info;
+    uint64_t hashTimes;
```
jserv, line 44: Avoid [camel case](https://en.wikipedia.org/wiki/Camel_case).

src/implcontext.c
```diff
@@ -64,3 +64,8 @@ int8_t *getPoWResult(ImplContext *impl_ctx, void *pow_ctx)
 {
     return impl_ctx->getPoWResult(pow_ctx);
 }
+
+uint64_t getHashTimes(ImplContext *impl_ctx, void *pow_ctx)
```
jserv, line 68: Why do you need implementation-specific counter?

src/pow_avx.c
```diff
@@ -681,6 +681,22 @@ static int8_t *PoWAVX_getPoWResult(void *pow_ctx)
     return ret;
 }
 
+static uint64_t PoWAVX_getHashTimes(void *pow_ctx)
```
jserv, line 684: There are various lines in common among `PoWAVX_getHashTimes`, `PoWC_getHashTimes`, etc. I don't think we really need that.

furuame: @marktwtn You can calculate the `HashTimes` inside the end of `doThePoW` function and store it. Then we can just use the interface get the `HashTimes` rather than duplicate `getHashTimes` functions in each implementation.

```
uint64_t getHashTimes(void *pow_ctx)
{
    return pow_ctx->hashtimes;
}
```

src/implcontext.c
furuame: We need a counter in each `pow_ctx`.

src/pow_sse.c
```diff
@@ -457,6 +457,22 @@ static int8_t *PoWSSE_getPoWResult(void *pow_ctx)
...
 
+static uint64_t PoWSSE_getHashTimes(void *pow_ctx)
+{
+    uint64_t count = 0;
+    PoW_SSE_Context *ctx = (PoW_SSE_Context *) pow_ctx;
+    Pwork_struct *pitem = ctx->pitem;
+
+    for (int i = 0; i < ctx->num_threads; i++) {
+        if (pitem[i].ret >= 0 )
+            count += (uint64_t)pitem[i].ret;
```
furuame, line 468: Please follow the coding style, thanks.

```
count += (uint64_t) pitem[i].ret;
```

src/pow_sse.c
```diff
@@ -457,6 +457,22 @@ static int8_t *PoWSSE_getPoWResult(void *pow_ctx)
...
+        else
+            count += (uint64_t)(-(pitem[i].ret) + 1);
```
furuame, line 470: Redundant parentheses

furuame (changes requested):

furuame:
`getHashTimes()` shoulde also be included in the test case.

src/clcontext.h
marktwtn: > getHashTimes() shoulde also be included in the test case.

Yes, it would be included in the future pull request of getting the hash rate.

marktwtn:
> getHashTimes() shoulde also be included in the test case.

Yes, it would be included in the other future pull request of getting the hash rate.

src/pow_avx.c
marktwtn: But the code would look like this
```
uint64_t getHashTimes(void *pow_ctx)
{
    return ((PoW_SSE_Context *) pow_ctx)->hashtimes;
}
```
and can not be used in other implementation.

I guess the `getHashTimes` function should remain the same and change the `PoW[Impl]_getHashTimes` function like this:

```
static uint64_t PoW[Impl]_getHashTimes(void *pow_ctx)
{
    return ((PoW_[Impl]_Context *) pow_ctx)->hashtimes;
}
```

src/implcontext.c
marktwtn: I'll set a counter named `hash_count` rather than `hash_times` to avoid misunderstanding in each `pow_ctx`.

And I'm not sure about the meaning of @jserv 's question.

jserv: I would like to see the methods to minimize changes against existing interface. We might split the functions of given interface into two fields:
* ops (operations)
* info (information)

That is, PoW takes ops after initialization, which expects info.

marktwtn:
I rewrite the code in #73 without implementing getHashTimes() API.

If there is no problem, I will close this pull request without merging.

src/pow_avx.c
furuame: You are right.

jserv:
> I rewrite the code in #73 without implementing getHashTimes() API.
> If there is no problem, I will close this pull request without merging.

Agree.

PR #73, opened 2018-09-21
Calculate and display hash count of PoW

marktwtn:
Calculate and sum the total `hash count` mentioned in #47 in `doThePoW` function.

The FPGA part will be implemented by @ajblane .

The `BUILD_BENCH` option and PoW execution time for hash rate will be added in the future pull requests.

2018/09/23: The `BUILD_BENCH` option has been added in `Makefile`.

### DIFF
--- a/src/pow_avx.c
+++ b/src/pow_avx.c
@@ -541,6 +541,7 @@ bool PowAVX(void *pow_ctx)
     /* Initialize the context */
     PoW_AVX_Context *ctx = (PoW_AVX_Context *) pow_ctx;
     ctx->stopPoW = 0;
+    ctx->hash_count = 0;
     pthread_mutex_init(&ctx->lock, NULL);
     pthread_t *threads = ctx->threads;
     Pwork_struct *pitem = ctx->pitem;
@@ -573,6 +574,7 @@ bool PowAVX(void *pow_ctx)
         pthread_join(threads[i], NULL);
         if (pitem[i].n == -1)
             completedIndex = i;
+        ctx->hash_count += (uint64_t) (pitem[i].ret >= 0 ? pitem[i].ret : -pitem[i].ret + 1);
     }
 
     nonce_trit = initTrits(nonce_array[completedIndex], NonceTrinarySize);
@@ -681,6 +683,11 @@ static int8_t *PoWAVX_getPoWResult(void *pow_ctx)
     return ret;
 }
 
+static uint64_t PoWAVX_getHashCount(void *pow_ctx)
+{
+    return ((PoW_AVX_Context *) pow_ctx)->hash_count;
+}
+
 ImplContext PoWAVX_Context = {
     .context = NULL,
     .description = "CPU (Intel AVX)",
@@ -693,4 +700,5 @@ ImplContext PoWAVX_Context = {
     .freePoWContext = PoWAVX_freePoWContext,
     .doThePoW = PowAVX,
     .getPoWResult = PoWAVX_getPoWResult,
+    .getHashCount = PoWAVX_getHashCount,
 };
--- a/src/pow_avx.c
+++ b/src/pow_avx.c
@@ -537,6 +537,7 @@ bool PowAVX(void *pow_ctx)
     bool res = true;
     Trits_t *nonce_trit = NULL;
     Trytes_t *tx_tryte = NULL, *nonce_tryte = NULL;
+    time_t start_time, end_time;
 
     /* Initialize the context */
     PoW_AVX_Context *ctx = (PoW_AVX_Context *) pow_ctx;
@@ -558,6 +559,7 @@ bool PowAVX(void *pow_ctx)
         goto fail;
     }
 
+    time(&start_time);
     /* Prepare arguments for pthread */
     for (int i = 0; i < ctx->num_threads; i++) {
         pitem[i].mid = c_state;
@@ -577,6 +579,8 @@ bool PowAVX(void *pow_ctx)
             completedIndex = i;
         ctx->pow_info->hash_count += (uint64_t) (pitem[i].ret >= 0 ? pitem[i].ret : -pitem[i].ret + 1);
     }
+    time(&end_time);
+    ctx->pow_info->time = difftime(end_time, start_time);
 
     nonce_trit = initTrits(nonce_array[completedIndex], NonceTrinarySize);
     if (!nonce_trit) {
--- a/src/pow_avx.c
+++ b/src/pow_avx.c
@@ -487,16 +487,16 @@ static int8_t *tx_to_cstate(Trytes_t *tx)
 {
     Trytes_t *inn = NULL;
     Trits_t *tr = NULL;
-    int8_t tyt[(transactionTrinarySize - HashSize) / 3] = {0};
+    int8_t tyt[TRANSACTION_TRYTES_LENGTH - HASH_TRYTES_LENGTH] = {0};
 
     Curl *c = initCurl();
     int8_t *c_state = (int8_t *) malloc(STATE_TRITS_LENGTH);
     if (!c || !c_state) goto fail;
 
-    /* Copy tx->data[:(transactionTrinarySize - HashSize) / 3] to tyt */
-    memcpy(tyt, tx->data, (transactionTrinarySize - HashSize) / 3);
+    /* Copy tx->data[:TRANSACTION_TRYTES_LENGTH - HASH_TRYTES_LENGTH] to tyt */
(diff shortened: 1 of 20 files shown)


furuame:
Why don't just integrate `BUILD_BENCH` option and measuring PoW execution time in this PR?

marktwtn:
> Why don't just integrate `BUILD_BENCH` option and measuring PoW execution time in this PR?

Just want to make sure the way of getting hash times is agreed by everyone.
I might be too careful.

src/pow_avx.c
```diff
@@ -589,6 +595,9 @@ bool PowAVX(void *pow_ctx)
 
     nonce_to_result(tx_tryte, nonce_tryte, ctx->output_trytes);
 
+    /* PoW-related information */
+    printf("Hash count: %"PRIu64"\n", ctx->hash_count);
```
jserv: Plain `printf` might not be informative to the glue code which utilizes dcurl such as IRI. You don't have to bother.

src/pow_avx.c
```diff
@@ -573,6 +575,10 @@ bool PowAVX(void *pow_ctx)
         pthread_join(threads[i], NULL);
         if (pitem[i].n == -1)
             completedIndex = i;
+        if (pitem[i].ret >= 0)
+            ctx->hash_count += (uint64_t) pitem[i].ret;
+        else
+            ctx->hash_count += (uint64_t) -pitem[i].ret + 1;
```
chenwei-tw: We can simplify it:
```
ctx->hash_count += (uint64_t) ((pitem[i].ret >= 0) ? pitem[i].ret : -pitem[i].ret + 1);
```

src/pow_avx.c
```diff
@@ -573,6 +575,10 @@ bool PowAVX(void *pow_ctx)
         pthread_join(threads[i], NULL);
         if (pitem[i].n == -1)
             completedIndex = i;
+        if (pitem[i].ret >= 0)
```
furuame: We can simplify it:
```
ctx->hash_count += (uint64_t) ((pitem[i].ret >= 0) ? pitem[i].ret : -pitem[i].ret + 1);
```

furuame (changes requested):

src/pow_avx.c
marktwtn: Done in the commit : https://github.com/DLTcollab/dcurl/pull/73/commits/370382269cbd3084df6669a9f97c4f18875c0726

src/pow_avx.c
marktwtn: The `printf` will be executed only when the `BUILD_BENCH=1` is specified.

See https://github.com/DLTcollab/dcurl/pull/73/commits/5b71fe3243b9dc9098d0d05f9cf97ee11c191172.

jserv:
I disagree with the way to place several `printf` inside PoW implementation. All core parts of dcurl should not contain verbose opertions over stdio. Instead, each function should only work as "black box" via certain interface specified by the public header(s). If users would like to benchmark, they can straightforwardly invoke utility function and manipulate stdio operations manually.

marktwtn:
I'll remove the `printf` inside PoW implementation.
It does make me feel weird but I just print it out without any doubt.

Back to the original implementation, I was going to create a new API to get the `hash times`.
Like the API of getting the PoW result.
I haven't figured out other better methods to get the `hash times`.
Is this the right way to do it?
I'm a little confused now. I might misunderstand some previous comments given to me.

jserv:
> Back to the original implementation, I was going to create a new API to get the `hash times`.
> Like the API of getting the PoW result.

It sounds fine to me.

> I haven't figured out other better methods to get the `hash times`.
> Is this the right way to do it?
> I'm a little confused now. I might misunderstand some previous comments given to me.

A Qt developer wrote the article [The Little Manual of API Design](https://people.mpi-inf.mpg.de/~jblanche/api-design.pdf) 10 years ago, and it is still insightful.

Quoted from 4.11 Strive for property-based APIs:
> Since the properties can be set in any order, it is usually necessary to use lazy initialization and other techniques in the implementation to avoid recomputing the whole object every time a property changes.

It is identically important to statistics in dcurl internals. To get statistics of hash rate, developers should be able to look up certain property via dedicated APIs. In this case, it would be great if there is something like bookkeeping as `info` in each implementation context.

src/pow_avx.c
marktwtn: The `printf` are removed from the `dcurl` core code.

marktwtn:
I use force push to overwrite the old commits with the new ones.

Now the `hash count` can be obtained with the `getHashCount` API.
The calculation and summation are done in the `doThePoW`.

It would show the hash count value when building and executing `test-pow.c` with `BUILD_BENCH` option enabled.

jserv:
Would it be meaingful if bench is replaced with stat, which stands for statistics and state?

marktwtn:
> Would it be meaingful if bench is replaced with stat, which stands for statistics and state?

I'm ok with both naming.
The naming `BUILD_BENCH` comes from the issue #47 .

@chenwei-tw what do you think?

furuame:
Fine for me 👍

furuame (approved):

chenwei-tw:
LGTM

furuame (changes requested):
Implement the API `getPoWInfo`

jserv:
@marktwtn, File `README.md` should be modified accordingly. Then, I would approve this pull request.

furuame:
There is no `getHashCount` API in the original `dev` branch or mentioned in other commit. This 7a916200 commit message is really weird.

tests/test-pow.c
```diff
@@ -112,6 +112,12 @@ int main()
         doThePoW(PoW_Context_ptr, pow_ctx);
         int8_t *ret_trytes = getPoWResult(PoW_Context_ptr, pow_ctx);
         assert(ret_trytes);
+#if defined(ENABLE_STAT)
+        PoW_Info *info = (PoW_Info *) getPoWInfo(PoW_Context_ptr, pow_ctx);
+        printf("Hash count: %"PRIu64"\n", info->hash_count);
+        printf("PoW execution time: %.0lf sec\n", info->time);
+        printf("Hash rate: %.3lf per sec\n", info->hash_count / info->time);
```
furuame: Use the "kH/sec".

jserv: According to [Wikipedia](https://en.wikipedia.org/wiki/Kilo-), Kilo is a decimal unit prefix in the metric system denoting multiplication by one thousand where it has the unit symbol k, in lower case.

tests/test-pow.c
```diff
@@ -112,6 +112,12 @@ int main()
         doThePoW(PoW_Context_ptr, pow_ctx);
         int8_t *ret_trytes = getPoWResult(PoW_Context_ptr, pow_ctx);
         assert(ret_trytes);
+#if defined(ENABLE_STAT)
```
furuame: I think the statistic should be printed after checking the PoW result is correct.

furuame (changes requested):

tests/test-pow.c
marktwtn: Due to the PoW information would be freed or destroyed after `destroyImplContext` function, I print the statistics in this order.

I'll declare variables to store the PoW information and print them after the verification.

src/pow_fpga_accel.h
```diff
@@ -13,6 +14,8 @@ struct _pow_fpga_accel_context {
     int8_t input_trytes[(transactionTrinarySize) / 3];  /* 2673 */
```
furuame: @marktwtn Could you help fix the marco?

marktwtn: Sure~

furuame:
Oops, hash count in fpga implementation has not been implemented yet.

ajblane:
I will send a pull request for hash count with the fpga accelerator.

ajblane:
There are compile errors for the fpga-accelerated PoW (!) and I will fix it.

jserv:
By the way, we shall set up CI for building all configurations including FPGA part.

ajblane:
There are also run-time errors the fpga-accelerated PoW.

marktwtn:
> There are compile errors for the fpga-accelerated PoW (!) and I will fix it.

That's my bad. I did not notice it.
The warning would disappear when enabling `BUILD_STAT` option.

> There are also run-time errors the fpga-accelerated PoW.

You can contact me on slack if you have any problem.

ajblane:
@marktwtn  Thanks.

PR #74, opened 2018-09-27
Fix bugs for fpga-accelerated PoW

ajblane:


furuame:
@ajblane @marktwtn  First, sorry for the bug which is created by me.

Then, is the hash count for fpga implementation included in this PR? If not, the commit a38f2695e06 should have better be included in the hash count PR .

ajblane:
> @ajblane @marktwtn First, sorry for the bug which is created by me.

@chenwei-tw  Forget it.

PR #75, opened 2018-09-27
Implement PoW info for the FPGA accelerator

ajblane:


marktwtn:
The commit 99244d4104a08b18358293112622a28061cc73b9 of the PR is the same as the commit db71be8cab6f2a0a21e9881d298a5b0fb0dd910c in `dev`.

Would you like to delete it by rebasing?

furuame:
The branch is in a mess. Please rebase the latest `dev` branch and remove the commit 
99244d4 and 02a2064

src/pow_fpga_accel.c
```diff
@@ -123,11 +146,37 @@ static bool PoWFPGAAccel_Context_Initialize(ImplContext *impl_ctx)
         goto fail_to_open_odata;
     }
 
+    ctx->pow_info = (PoW_Info *) malloc(sizeof(PoW_Info));
```
furuame: If the fpga implementation is default for one thread, the `PoW_Info` member can be declared as object rather than pointer and we can avoid this malloc.

src/pow_fpga_accel.c
```diff
@@ -145,6 +194,13 @@ static void PoWFPGAAccel_Context_Destroy(ImplContext *impl_ctx)
     close(ctx->out_fd);
     close(ctx->ctrl_fd);
 
+    int result = munmap(ctx->fpga_regs_map, HPS_TO_FPGA_SPAN);
+    if (result < 0) {
+        perror("devmem munmap fail");
+    }
+    close(ctx->devmem_fd);
+
+    free(ctx->pow_info);
```
furuame: ditto.

src/pow_fpga_accel.c
furuame: If so, the implementation of `PoWFPGAAccel_getPoWInfo` is also needed to modify.

PR #79, opened 2018-10-04
Print the range including 95% of the hash rate values

### LINKED ISSUE #77: Acquire reliable hash rate distribution with the method of Statistics
The hash rate would be different at each time we execute PoW with the same hardware.
There might be a lot of factors which influence the hash rate value.

We try to measure the hash rate distribution and remove the extremely large and small hash rate.
A reliable range of hash rate value should be got with the method of Statistics.

marktwtn:
Implement functions of getting the average value and standard deviation.
Print the hash rate range when the `BUILD_STAT` option is enabled.
Close #77.

-----

The output message with `BUILD_STAT` option enabled would look like this:
```
Hash rate average value: XXXX.XXX kH/sec
with the range +- xxx.xxx kH/sec including 95% of the hash rate values.
```

### DIFF
--- a/tests/test-pow.c
+++ b/tests/test-pow.c
@@ -1,6 +1,9 @@
 /* Test program for pow_*.c */
 #include "common.h"
 #include "implcontext.h"
+#include "math.h"
+
+#define POW_TOTAL 100
 
 #if defined(ENABLE_AVX)
 extern ImplContext PoWAVX_Context;
@@ -35,6 +38,29 @@ const char *description[ ] = {
     "FPGA",
 #endif
 };
+
+double getAvg(const double arr[], int arrLen) {
+    double avg, sum = 0;
+
+    for (int idx = 0; idx < arrLen; idx++) {
+        sum += arr[idx];
+    }
+    avg = sum / arrLen;
+
+    return avg;
+}
+
+double getStdDeviation(const double arr[], int arrLen) {
+    double sigma, variance = 0;
+    double avg = getAvg(arr, arrLen);
+
+    for (int idx = 0; idx < arrLen; idx++) {
+        variance += pow(arr[idx] - avg, 2);
+    }
+    sigma = sqrt(variance / arrLen);
+
+    return sigma;
+}
 
 int main()
 {
@@ -98,9 +124,16 @@ int main()
         PoWFPGAAccel_Context,
 #endif
     };
+#if defined(ENABLE_STAT)
+    double hashRateArr[POW_TOTAL];
+    int pow_total = POW_TOTAL;
+#else
+    int pow_total = 1;
+#endif
+
 
     for (int idx = 0; idx < sizeof(ImplContextArr) / sizeof(ImplContext); idx++) {
-        printf("%s\n",description[idx]);
+        printf("%s\n", description[idx]);
 
         ImplContext *PoW_Context_ptr = &ImplContextArr[idx];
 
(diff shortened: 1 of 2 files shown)


furuame:
Here is my suggestion:
1. How about executing PoW `POW_TOTAL` times only when `ENABLE_STAT` enabled? We just need to check correctness unless `ENABLE_STATE` enabled.
2. I'm not sure whether executing the PoW for the "same" transaction tryte many times achieve the original goal. Or, you can also consider doing PoW for different transaction trytes. There are some testing data in `tests/trytes.txt`

marktwtn:
>     1. How about executing PoW `POW_TOTAL` times only when `ENABLE_STAT` enabled?
>     We just need to check correctness unless `ENABLE_STATE` enabled.
>

I agree. Just make it simple.
If `ENABLE_STATE` is disabled, check the correctness.
If `ENABLE_STATE` is enabled, check the correctness and do the hash rate statistics analysis.

The program structure would be the same, except that there would be a variable which controls the total execution times of PoW.
If `ENABLE_STATE` is disabled, the total execution times equals to 1.
If `ENABLE_STATE` is enabled, the total execution times equals to macro `POW_TOTAL`.

>     2. I'm not sure whether executing the PoW for the "same" transaction tryte many times
>     achieve the original goal.
>     Or, you can also consider doing PoW for different transaction trytes.
>     There are some testing data in `tests/trytes.txt`

I guess it still achieves the original goal since we only care about the hash rate, which is dependent to the implementation?

furuame:
Don't you care the elapsed time of each PoW?

marktwtn:
We only focus on the hash rate now.
If the elapsed time of PoW of different transaction tryte needs to be shown, I will make sure each execution of PoW uses the different transaction tryte.

tests/test-pow.c
```diff
@@ -98,44 +124,59 @@ int main()
...
-        for (int i = HASH_TRITS_LENGTH - 1; i >= HASH_TRITS_LENGTH - mwm; i--) {
-            assert(ret_trits->data[i] == 0);
-        }
+        printf("%s\n", description[idx]);
+
+        for (int count = 0; count < pow_total; count++) {
+            ImplContext *PoW_Context_ptr = &ImplContextArr[idx];
+
+            /* test implementation with mwm = 14 */
+            initializeImplContext(PoW_Context_ptr);
```
furuame: Just make sure that is it duplicate initialization of `ImplContext` necessary?

tests/test-pow.c
```diff
@@ -98,44 +124,59 @@ int main()
...
+            void *pow_ctx = getPoWContext(PoW_Context_ptr, (int8_t *) trytes, mwm);
+            assert(pow_ctx);
+            doThePoW(PoW_Context_ptr, pow_ctx);
+            int8_t *ret_trytes = getPoWResult(PoW_Context_ptr, pow_ctx);
+            assert(ret_trytes);
+#if defined(ENABLE_STAT)
+            PoW_Info pow_info = getPoWInfo(PoW_Context_ptr, pow_ctx);
+#endif
+            freePoWContext(PoW_Context_ptr, pow_ctx);
+            destroyImplContext(PoW_Context_ptr);
```
furuame: ditto.

tests/test-pow.c
furuame: It would be initialized 100 times in this loop.

marktwtn: It is not necessary and I will rewrite it.
The original loop design is based on the previous discussion that the PoW information should be got after the validation.

Therefore, I will reorder the code which makes `freePoWContext` and `destroyImplContext` functions execute after the validation.
The loop executing `n(=100)` times would contain the code being related to `PoW (execution, getting result and information)` and `validation`.

tests/test-pow.c
```diff
@@ -98,44 +124,61 @@ int main()
...
+            /* Validation */
+            for (int i = HASH_TRITS_LENGTH - 1; i >= HASH_TRITS_LENGTH - mwm; i--) {
+                assert(ret_trits->data[i] == 0);
+            }
+
+            free(ret_trytes);
+            freeTrobject(trytes_t);
+            freeTrobject(hash_trytes);
+            freeTrobject(ret_trits);
+
```
furuame, line 169: Though the `test-pow` only be used to test one implementation, the `freePoWContext` and `destroyImplContext` should be placed here beacause there is a outer loop to iterate different `ImplContext`.

marktwtn: Mmm...... I do not see any problem here.

PR #87, opened 2018-11-27
Support "--pow-threads" command line option of IRI

### LINKED ISSUE #86: Support the new command line option "--pow-threads" of IRI in dcurl
The IOTA `IRI` [commit](https://github.com/iotaledger/iri/commit/51ebe60a3a80da9885bd1653e4c8f47c550df798) has added the new command line option `--pow-threads` for assigning **Number of threads to use for proof-of-work calculation**.

The `dcurl` should support the `--pow-threads` in CPU implementation.

The priority of the CPU thread number would be:
1. Assigned by **IRI** `--pow-threads`
2. Environment variable `DCURL_NUM_CPU`
3. Core number detected by **dcurl**

marktwtn:
If the "--pow-threads" value is greater than 0 and less than the maximum available resources,
the number of threads would be equal to the "--pow-threads" value.
Otherwise, the number of threads would be set with the original mechanism.

The GPU and FPGA implementation are not affected by the "--pow-threads".

Close #86.

### DIFF
--- a/jni/iri-pearldiver-exlib.c
+++ b/jni/iri-pearldiver-exlib.c
@@ -16,7 +16,8 @@ JNIEXPORT jboolean JNICALL
 Java_com_iota_iri_hash_PearlDiver_exlib_1search(JNIEnv *env,
                                                 jclass clazz,
                                                 jbyteArray trits,
-                                                jint mwm)
+                                                jint mwm,
+                                                jint threads)
 {
     /*********** Get the Byte array from Java byte Array *************/
     jbyte *c_trits = (*env)->GetByteArrayElements(env, trits, NULL);
@@ -27,7 +28,7 @@ Java_com_iota_iri_hash_PearlDiver_exlib_1search(JNIEnv *env,
         return JNI_FALSE;
     /****************************************************************/
 
-    int8_t *result = dcurl_entry(arg_trytes->data, mwm);
+    int8_t *result = dcurl_entry(arg_trytes->data, mwm, threads);
 
     /************ Write result back Java byte array *****************/
     Trytes_t *ret_trytes = initTrytes(result, 2673);
--- a/src/compat-ccurl.c
+++ b/src/compat-ccurl.c
@@ -21,7 +21,7 @@ char *ccurl_pow(char *trytes, int mwm)
         isInitialized = true;
     }
     pthread_mutex_unlock(&mtx);
-    return (char *) dcurl_entry((int8_t *) trytes, mwm);
+    return (char *) dcurl_entry((int8_t *) trytes, mwm, 1);
 }
 
 void ccurl_pow_finalize(void)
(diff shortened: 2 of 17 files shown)


src/pow_c.c
```diff
@@ -374,6 +374,8 @@ static void *PoWC_getPoWContext(ImplContext *impl_ctx, int8_t *trytes, int mwm)
             memcpy(ctx->input_trytes, trytes, TRANSACTION_TRYTES_LENGTH);
             ctx->mwm = mwm;
             ctx->indexOfContext = i;
+            if (threads > 0 && threads < ctx->num_threads)
+                ctx->num_threads = threads;
```
furuame, line 378: Should we reset `ctx->num_threads` when this PoW Context is freed?

marktwtn: The scenario I was thinking is using `IRI` + `dcurl`.
Since the `--pow-threads` is assigned when activating the `IRI`,
the number of threads would be fixed during the execution time.
It would not have any problem.

However, if we use `dcurl` alone, it may come up with issues.
For example:
```
getPoWContext with thread number = N
...
freePoWContext
...
getPoWContext with thread number = N+1
```

I would like to reset `ctx->num_threads` by the pre-stored maximum thread value.
The stored value would be the new member in `struct _pow_XXX_context` called `num_max_threads`.
Or just call the function `get_avail_nprocs()` to reset it.

marktwtn: No reset when the PoW Context is freed.
Just make sure the assigned thread number is in the reasonable range.

src/compat-ccurl.c
```diff
@@ -21,7 +21,7 @@ char *ccurl_pow(char *trytes, int mwm)
         isInitialized = true;
     }
     pthread_mutex_unlock(&mtx);
-    return (char *) dcurl_entry((int8_t *) trytes, mwm);
+    return (char *) dcurl_entry((int8_t *) trytes, mwm, 0);
```
jserv: I have no idea about the reason why it is defaulted to `0` instead of `1`?

marktwtn: Oops, my bad.

It has been changed to `1`.

jni/iri-pearldiver-exlib.c
```diff
@@ -27,7 +28,7 @@ Java_com_iota_iri_hash_PearlDiver_exlib_1search(JNIEnv *env,
         return JNI_FALSE;
     /****************************************************************/
 
-    int8_t *result = dcurl_entry(arg_trytes->data, mwm);
+    int8_t *result = dcurl_entry(arg_trytes->data, mwm, threads);
```
jserv, line 31: What if `threads` equals to `0`?

jserv: Will IRI be functioned unexpectedly?

marktwtn: Nope.

If the `--pow-threads` option is not used, the default value of **IRI** thread number would be `0`.
In that case, the available thread number would be automatically detected by **dcurl**.

Possible value:
- Max core number - 1
- Environment variable `DCURL_NUM_CPU` ( <= max core number - 1 )

marktwtn: However, I forgot to assign the value detected by `dcurl`.

marktwtn: Done.

PR #93, opened 2019-01-30
Integrate libtuv thread pool to eliminate thread creation overhead

### LINKED ISSUE #58: Facilitate thread pool to eliminate overhead while creating threads frequently
Since every CPU implementations, such as SSE and AVX, create many threads to find the nonce every time, the overhead cannot be ignored. [`libtuv`](https://github.com/Samsung/libtuv/blob/master/src/threadpool.c) offers a lightweight threadpool implementation for the embedded system.

This feature is expected to be implemented after the issue #57 is achieved. When the new interface is applied in `ducrl`, the threadpool can be hidden in the CPU PoW Algorithm context.

marktwtn:
To reduce the overhead of creating and eliminating the threads repeatedly,
we integrate the thread pool of libtuv with git submodule.
The pthread-related functions and data types are replaced with the corresonding
ones of libtuv.
The compilation of libtuv library is written in the file mk/submodule.mk.

Experiment:
Call clock_gettime() right before and after the functions for getting the thread.
The functions are pthread_create() (without thread pool)
and uv_queue_work() (with thread pool).
Use test-multi-pow.py as testcase since it initializes and destroys dcurl only once and
does the PoW multiple times, like what IRI does.
The experiment result shows the time of getting each thread
and the thread number of a PoW execution is 7.

Experiment result (unit: second):
Without thread pool
thread0: 0.000028384
thread1: 0.000025127
thread2: 0.000024748
thread3: 0.000023925
thread4: 0.000024126
thread5: 0.000025328
thread6: 0.000052900
thread0: 0.000049344
thread1: 0.000039575
thread2: 0.000036720
thread3: 0.000036249
thread4: 0.000034606
thread5: 0.000034676
thread6: 0.000033444

With thread pool
thread0: 0.000124327
thread1: 0.000002084
thread2: 0.000001052
thread3: 0.000000150
thread4: 0.000000121
thread5: 0.000000080
thread6: 0.000000090
thread0: 0.000000291
thread1: 0.000000080
thread2: 0.000000050
thread3: 0.000000050
thread4: 0.000000050
thread5: 0.000000060
thread6: 0.000000050

The first consumed time of getting the thread from thread pool is longer
since it is in charge of preallocating and initalizing the threads.

Close #58.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,9 @@ endif
 # Check specific CPU features available on build host
 include mk/cpu-features.mk
 
+# Handle git submodule
+include mk/submodule.mk
+
 ifeq ("$(BUILD_AVX)","1")
 CFLAGS += -mavx -DENABLE_AVX
 ifeq ("$(call cpu_feature,AVX2)","1")
@@ -117,17 +120,17 @@ OBJS := $(addprefix $(OUT)/, $(OBJS))
 
 $(OUT)/test-%.o: tests/test-%.c
 	$(VECHO) "  CC\t$@\n"
-	$(Q)$(CC) -o $@ $(CFLAGS) -I $(SRC) -c -MMD -MF $@.d $<
+	$(Q)$(CC) -o $@ $(CFLAGS) -I $(SRC) $(LIBTUV_INCLUDE) -c -MMD -MF $@.d $<
 
 $(OUT)/%.o: $(SRC)/%.c
 	$(VECHO) "  CC\t$@\n"
-	$(Q)$(CC) -o $@ $(CFLAGS) -c -MMD -MF $@.d $<
+	$(Q)$(CC) -o $@ $(CFLAGS) $(LIBTUV_INCLUDE) -c -MMD -MF $@.d $<
 
-$(OUT)/test-%: $(OUT)/test-%.o $(OBJS)
+$(OUT)/test-%: $(OUT)/test-%.o $(OBJS) $(LIBTUV_LIBRARY)
 	$(VECHO) "  LD\t$@\n"
 	$(Q)$(CC) -o $@ $^ $(LDFLAGS)
 
-$(OUT)/libdcurl.so: $(OBJS)
+$(OUT)/libdcurl.so: $(OBJS) $(LIBTUV_LIBRARY)
 	$(VECHO) "  LD\t$@\n"
 	$(Q)$(CC) -shared -o $@ $^ $(LDFLAGS)
 
--- a/mk/submodule.mk
+++ b/mk/submodule.mk
@@ -1,0 +1,9 @@
+# libtuv related variables
+LIBTUV_PATH = deps/libtuv
+LIBTUV_INCLUDE := -I $(LIBTUV_PATH)/include
+# PIC (Position-Independent-Code) library
+LIBTUV_LIBRARY := $(LIBTUV_PATH)/build/x86_64-linux/release/lib/libtuv.o
+
+$(LIBTUV_LIBRARY):
+	git submodule update --init $(LIBTUV_PATH)
+	$(MAKE) -C $(LIBTUV_PATH) TUV_BUILD_TYPE=release TUV_CREATE_PIC_LIB=yes
--- a/mk/submodule.mk
+++ b/mk/submodule.mk
@@ -1,8 +1,20 @@
+# Copy from the Makefile of libtuv to support different platforms
+UNAME_M := $(shell uname -m)
+UNAME_S := $(shell uname -s)
+ifeq ($(UNAME_S),Linux)
+    UNAME_S := linux
+endif
+ifeq ($(UNAME_S),Darwin)
+    # macOS
+    UNAME_S := darwin
+endif
+
 # libtuv related variables
 LIBTUV_PATH = deps/libtuv
 LIBTUV_INCLUDE := -I $(LIBTUV_PATH)/include
+LIBTUV_PLATFORM := $(UNAME_M)-$(UNAME_S)
 # PIC (Position-Independent-Code) library
-LIBTUV_LIBRARY := $(LIBTUV_PATH)/build/x86_64-linux/release/lib/libtuv.o
+LIBTUV_LIBRARY := $(LIBTUV_PATH)/build/$(LIBTUV_PLATFORM)/release/lib/libtuv.o
 
 $(LIBTUV_LIBRARY):
 	git submodule update --init $(LIBTUV_PATH)
--- a/src/pow_c.c
+++ b/src/pow_c.c
@@ -10,6 +10,7 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <string.h>
+#include <uv.h>
 #include "cpu-utils.h"
 #include "curl.h"
 #include "implcontext.h"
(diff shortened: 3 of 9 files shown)


jserv:
The commit messages shall contain relevant experiment results for reference purpose.

README.md
```diff
@@ -11,9 +11,10 @@ enabled in multi-threaded execuction fashion, resulting in much faster proof-of-
 Reference Implementation (IRI). Additionally, dcurl also supports the FPGA-accelerated solution further described in docs/FPGA-ACCEL.md
 
 # Warning
-* You need to configure paths and flags of OpenCL installation in ```mk/opencl.mk```
+* You need to configure paths and flags of OpenCL installation in ```mk/opencl.mk```.
 * dcurl will automatically configure all the GPU divices on your platform.
 * Check JDK installation and set JAVA_HOME if you wish to specify.
+* Initialize and update git submodules with `$ git submodule init` and `$ git submodule update` after downloading the `dcurl` repository.
```
jserv: Don't mention this. Build system proceeds instead.

Makefile
```diff
@@ -27,6 +27,9 @@ endif
 # Check specific CPU features available on build host
 include mk/cpu-features.mk
 
+# Handle git submodule
+include mk/submodule.mk
```
wusyong, line 31: How about init and update submodule automatically in during `make` like [accelerator does](https://github.com/DLTcollab/tangle-accelerator/blob/master/Makefile#L12)?

jserv:
For performance measurement, you should take [C++11 threads, affinity and hyperthreading](https://eli.thegreenplace.net/2016/c11-threads-affinity-and-hyperthreading/) into considerations. Thread affinity might differentiate dramatically.

src/pow_c.c
```diff
@@ -190,7 +191,10 @@ static void *pworkThread(void *pitem)
         pworkInfo->n = -1;
     }
     pthread_mutex_unlock(pworkInfo->lock);
-    pthread_exit(NULL);
+}
+
+static void work_cb(uv_work_t *req) {
+    pworkThread(req->data);
```
jserv: It is not necessary to keep function `pworkThread`. Instead, you can implement the existing logic in function `work_cb`.

jserv:
How about C and OpenCL implementations?

Makefile
marktwtn: Thanks for suggestion.
I will initialize and update submodule automatically in `mk/submodule.mk`.

src/pow_c.c
marktwtn: The function `pworkThread` has been replaced with function `work_cb`.

README.md
marktwtn: Done.

marktwtn:
> The commit messages shall contain relevant experiment results for reference purpose.

The experiment result has been added/updated to the git commit and the first comment of this issue.

---

> For performance measurement, you should take [C++11 threads, affinity and hyperthreading](https://eli.thegreenplace.net/2016/c11-threads-affinity-and-hyperthreading/) into considerations. Thread affinity might differentiate dramatically.

I think this can be opened as another issue?

---



> How about C and OpenCL implementations?

C implementation does include thread pool of libtuv.
OpenCL implementation seems nothing to do with the thread?

jserv:
> > For performance measurement, you should take [C++11 threads, affinity and hyperthreading](https://eli.thegreenplace.net/2016/c11-threads-affinity-and-hyperthreading/) into considerations. Thread affinity might differentiate dramatically.
> 
> I think this can be opened as another issue?

Yes, please do. The skeleton implementation looks like the following:
```c
static int num_processors;

int main(int argc, char *argv[]) {
    ...
    num_processors = sysconf(_SC_NPROCESSORS_CONF);
    ...
}

#include <sched.h>
static inline void drop_policy(void) {
    struct sched_param param = { .sched_priority = 0;  };
    sched_setscheduler(0, SCHED_OTHER, &param);
}

static inline void affine_to_cpu(int id, int cpu) {
    cpu_set_t set;
    CPU_ZERO(&set);
    CPU_SET(cpu, &set);
    sched_setaffinity(0, sizeof(set), &set);
}

static void *worker_thread(void *userdata) {
    int thread_id = ((thread_info *) userdata)->id;
    ...
    /* Set worker threads to nice 19 and then preferentially to SCHED_IDLE
     * and if that fails, then SCHED_BATCH. No need for this to be an
     * error if it fails.
     */
    if (!geteuid())
        setpriority(PRIO_PROCESS, 0, -14);
    drop_policy();

    /* Cpu affinity only makes sense if the number of threads is a multiple
     * of the number of CPUs.
     */
    affine_to_cpu(thread_id, thread_id % num_processors);
    ...
}
```

mk/submodule.mk
```diff
@@ -0,0 +1,9 @@
+# libtuv related variables
+LIBTUV_PATH = deps/libtuv
+LIBTUV_INCLUDE_PATH := $(LIBTUV_PATH)/include
```
jserv: Change from `LIBTUV_INCLUDE_PATH` to `LIBTUV_INCLUDE` and re-assign it to `-I $(LIBTUV_PATH)/include`.

jserv:
> OpenCL implementation seems nothing to do with the thread?

(off-topic) Is `pthread_mutex_lock` necessary in OpenCL backend? Can we simply [synchronize at a barrier](https://linux.die.net/man/3/pthread_barrier_init)?

mk/submodule.mk
jserv: Alternatively, we can follow the naming convention of [pkg-config](https://www.freedesktop.org/wiki/Software/pkg-config/) to rename libtuv-specific variables as `LIBTUV_CFLAGS` and `LIBTUV_LIBS`.

jserv:
Experiment result shall come with the listing of hardware configurations for reference purpose.

jserv:
Since this pull request dramatically changes the flow of execution, there should be a dedicated note briefing fundamental designs in directory `docs/`.  Something like `docs/threading-model.md` would be nice where we can discuss thread pool, SMP affinity, load balancing, etc.

mk/submodule.mk
marktwtn: I am curious about the benefit of this writing skill.

jserv: No visible benefit but consistency. The reason why `LIBTUV_INCLUDE` or `LIBTUV_CFLAGS` should contain `-I` prefix is to reorder search paths if necessary.

jserv:
Rebasing is required due to recent document re-organization.

mk/submodule.mk
```diff
@@ -0,0 +1,9 @@
...
+# PIC (Position-Independent-Code) library
+LIBTUV_LIBRARY := $(LIBTUV_PATH)/build/x86_64-linux/release/lib/libtuv.o
```
jserv: The generation of `x86_64-linux` is a bit confusing since dcurl itself is portable. Both macOS and GNU/Linux are supported, not to mention arm, aarch64, and x86_64 architecture. We should avoid such hardcoded path though.

marktwtn: `libtuv` supports different platforms.
I will copy the corresponding part of `Makefile` in `libtuv` to avoid hardcoded path.

marktwtn: It has been fixed in the new commit.

mk/submodule.mk
marktwtn: Done.

mk/submodule.mk
```diff
@@ -0,0 +1,21 @@
+# Copy from the Makefile of libtuv to support different platforms
+UNAME_M := $(shell uname -m)
+UNAME_S := $(shell uname -s)
```
jserv, line 3: Add `FIXME` to mention the limitation of supported operating system listing.

marktwtn:
> Experiment result shall come with the listing of hardware configurations for reference purpose.

I have added the architecture and CPU information in the commit message.

---

> Since this pull request dramatically changes the flow of execution, there should be a dedicated note briefing fundamental designs in directory `docs/`. Something like `docs/threading-model.md` would be nice where we can discuss thread pool, SMP affinity, load balancing, etc.

I will record it as a TO-DO list.

---

> Rebasing is required due to recent document re-organization.

This has been done without any problem.

marktwtn:
> > OpenCL implementation seems nothing to do with the thread?
> 
> (off-topic) Is `pthread_mutex_lock` necessary in OpenCL backend? Can we simply [synchronize at a barrier](https://linux.die.net/man/3/pthread_barrier_init)?

It is necessary.
It makes sure that we can select the right GPU hardware if we have more than one GPU device.

mk/submodule.mk
marktwtn: Umm......
I'm not sure what you expect to see.
Like listing the operating system that `dcurl` supports but `libtuv` does not or vice versa?

PR #104, opened 2019-02-14
Messages dumped by printf should be only valid for debug build

### LINKED ISSUE #103: No printf should be used for release build
The intension of `dcurl` is to accelerate PoW, and its execution is almost hidden from the perspective of deployment. Therefore, any message dumped by `printf` are useless for release build.

The message `[dcurl] Implementation CPU (Intel AVX) is initialized successfully` should only appear in debug build.

jserv:
Close #103

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ endif
 -include $(OUT)/local.mk
 
 ifeq ("$(BUILD_DEBUG)","1")
-CFLAGS += -Og -g3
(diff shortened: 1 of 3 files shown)


marktwtn:
The `docs/build-n-test.md` should be modified as well.

The **Expected Results** would be different,
and I found out that we do not have `BUILD_DEBUG` explanation in **Build Instructions**.

PR #115, opened 2019-02-26
docs: add the comment and version of dcurl API

marktwtn:
Add the comment and configuration file of Doxygen to generate the API document.
The README.md also adds a badge of the github release version.
At last, the new macros for dcurl version are added.

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -1,0 +1,32 @@
+#---------------------------------------------------------------------------
+# Project related configuration options
+#---------------------------------------------------------------------------
+PROJECT_NAME           = "dcurl"
+PROJECT_NUMBER         = 0.1.0
+OUTPUT_DIRECTORY       = docs
+OPTIMIZE_OUTPUT_FOR_C  = YES
+#---------------------------------------------------------------------------
+# Configuration options related to warning and progress messages
+#---------------------------------------------------------------------------
+QUIET                  = YES
+#---------------------------------------------------------------------------
+# Configuration options related to the input files
+#---------------------------------------------------------------------------
+INPUT                  = . \
+                         docs \
+                         src
+FILE_PATTERNS          = dcurl.c \
+                         dcurl.h \
+                         *.md
+USE_MDFILE_AS_MAINPAGE = README.md
+#---------------------------------------------------------------------------
+# Configuration options related to source browsing
+#---------------------------------------------------------------------------
+SOURCE_BROWSER         = YES
+INLINE_SOURCES         = YES
+REFERENCED_BY_RELATION = YES
+REFERENCES_RELATION    = YES
+#---------------------------------------------------------------------------
+# Configuration options related to the LaTeX output
+#---------------------------------------------------------------------------
+GENERATE_LATEX         = NO
--- a/src/dcurl.h
+++ b/src/dcurl.h
@@ -4,8 +4,51 @@
 #include <stdbool.h>
 #include "trinary.h"
 
+/**
+ * @file dcurl.h
+ * @brief dcurl API functions.
+ *
+ * dcurl is a hardware-accelerated implementation for IOTA PearlDiver.
+ * It supports multi-threaded SIMD on CPU, OpenCL on GPU and FPGA to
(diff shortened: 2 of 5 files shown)


Doxyfile
```diff
@@ -0,0 +1,38 @@
...
+#---------------------------------------------------------------------------
+QUIET                  = YES
+#---------------------------------------------------------------------------
+# Configuration options related to the input files
+#---------------------------------------------------------------------------
+INPUT                  = . \
+                         docs \
+                         src
+FILE_PATTERNS          = dcurl.c \
+                         dcurl.h \
```
wusyong, line 19: I think documentation could just all put in dcurl.h

marktwtn: I searched on the Internet and found out that writing the document in C source file or header file is acceptable.
It only suggests that we should not write the document on both side.

The reason I wrote in source file is if the functions are changed, they can be modified without opening another file.

I think no matter which way we choose, just make sure every project follows the same rule.

wusyong (approved):

src/dcurl.c
```diff
@@ -99,6 +117,17 @@ void dcurl_destroy()
...
 
 
+/**
+ * @brief dcurl execution.
+ *
+ * Retrieve the available hardware from the list and use it to do the
+ * PoW(Proof-of-Work).
+ * @param [in] trytes The trytes for PoW calculation.
+ * @param [in] mwm The minimum weight magnitude.
+ * @param [in] threads The thread number of calculating the PoW. It affects CPU
```
wusyong: Detailed explanation of thread amount may needed (like 0 is actually max amount - 1.)

Doxyfile
marktwtn: The comment has been moved to the header file.

src/dcurl.c
marktwtn: The explanation of threads has been updated.

Doxyfile
marktwtn: I did not remove the `dcurl.c` in `FILE_PATTERNS` since I want to make the users to link and check the source code easily.

src/dcurl.c
```diff
@@ -9,11 +9,10 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
-#if defined(ENABLE_OPENCL)
-#include "pow_cl.h"
-#endif
-#if defined(ENABLE_FPGA_ACCEL)
-#include "pow_fpga_accel.h"
+#ifdef __APPLE__
```
jserv: Why did you change `src/dcurl.c`?

marktwtn: The original `Doxygen` comment was written in the `dcurl.c`.
I also changed the order of the header files for better classification.

Since the comment has been moved to `dcurl.h`, I just left the rest of the modification in the commit.

marktwtn: Shall I revert `dcurl.c` for a cleaner commit?

jserv: Yes, please do. Only preserve functional changes.

src/dcurl.h
```diff
@@ -4,8 +4,44 @@
 #include <stdbool.h>
 #include "trinary.h"
 
+/**
+ * @file dcurl.h
+ */
```
jserv, line 16: Add general descriptions about dcurl here.

src/dcurl.c
marktwtn: Done.

src/dcurl.h
marktwtn: Done.

marktwtn: Did you mean the general descriptions about dcurl rather than the description of this file?
I think I might update the comment with the latter one.

jserv: It would be something describing the functionality about dcurl.
> dcurl library is provided as hardware-accelerated implementation for IOTA PearlDiver, which utilizes multi-threaded SIMD, FPGA and GPU.

marktwtn: Done.

Issue #91, opened 2019-01-25
Remote PoW service to manipulate queued requests


jserv:
At the moment, `dcurl` already provides preliminary FPGA acceleration with consistent APIs, which enable transparent switching among SIMD, OpenCL, and FPGA implementations. However, it is not straightforward to accelerate existing IRI instances with FPGA accelerations because of the interoperability. Thus, we shall provide a remote PoW service dedicated to queued requests as [PoWbox](https://blog.iota.org/relaunching-the-powbox-d392236b6939) does.

Expected output:
1. Use existing libraries, such as ZeroMQ and ProtocolBuffer, to marshal / serialize the PoW operations;
2. Provide client-side interface to access FPGA-accelerated machines. Cyclone V-based device is the preferable target;
3. Verify IRI or Tangle-Accelerator + dcurl integration with FPGA accelerations;

Reference:
1. [PoWbox](https://github.com/iotaledger/powbox)
2. [diverDriver](https://github.com/muXxer/diverDriver)

tylerw1369:
If you need help testing let me know I can on my nano.  I can even allow you access to my nano if needed

jserv:
However, I am a bit afraid that PoWbox might be too heavy to launch on devices powered by Cyclone V. [powsrv](https://gitlab.com/powsrv.io) just implements slim interface with protocol buffers.

ajblane:
client-side interface is expected as follows:

* remote_interface_init()
check rabbitmq broker is alive

* remote_interface_create_pow_work()
asynchronously send transaction and mwm to rabbitmq broker

* remotre_interface_get_pow_result()
asynchronously get result from rabbitmq broker

* remote_interface_destroy()
leave rabbitmq broker

Issues:
1. When client-side requests are too many, we can build and manage more than one rabbitmq broker.
2. Need test bulk transfer effort

ajblane:
For managing RabbitMQ broker, RabbitMQ provides management plugin, names [rabbitmq_management](https://www.rabbitmq.com/management.html). It's very friendly and can show internal information. We need to be familiar with [rabbitmq_management](https://www.cloudamqp.com/blog/2015-05-27-part3-rabbitmq-for-beginners_the-management-interface.html) to manage it.

ajblane:
> However, I am a bit afraid that PoWbox might be too heavy to launch on devices powered by Cyclone V. powsrv just implements slim interface with protocol buffers.

We only require RabbitMQ C client library on devices.

[...]

PR #117, opened 2019-03-02
feat: Remote interface interoperate distributed computing resources

ajblane:
* Make dcurl surpport asynchronous remote procedure call
* Use RabbitMQ broker to manage queues
* Implement dcurl worker
* Support a scenario of a broker and many dcurl workers
* How to test remote dcurl, see docs/remote-dcurl.md


.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "deps/libtuv"]
 	path = deps/libtuv
 	url = https://github.com/DLTcollab/libtuv.git
+[submodule "deps/rabbitmq-c"]
+	path = deps/rabbitmq-c
+	url = https://github.com/DLTcollab/rabbitmq-c.git
```
jserv: Since there is no visible change against `rabbitmq-c`, why did you fork from RabbitMQ C client?

jserv:
The feature of remote interface should be optional, that implies extra configuration(s).

Makefile
```diff
@@ -137,17 +142,17 @@ ifeq ("$(ENABLE_CPU_PLATFORMS)","1")
...
+	$(Q)$(CC) -o $@ $(CFLAGS) -I $(SRC) $(LIBTUV_INCLUDE) $(LIBRABBITMQ_INCLUDE) -c -MMD -MF $@.d $<
 
 $(OUT)/%.o: $(SRC)/%.c $(LIBTUV_PATH)/include
 	$(VECHO) "  CC\t$@\n"
 	$(Q)$(CC) -o $@ $(CFLAGS) $(LIBTUV_INCLUDE) -c -MMD -MF $@.d $<
 
 $(OUT)/test-%: $(OUT)/test-%.o $(OBJS)
 	$(VECHO) "  LD\t$@\n"
-	$(Q)$(CC) -o $@ $^ $(LDFLAGS)
+	$(Q)$(CC) -o $@ $^ $(LDFLAGS) $(LIBRABBITMQ_LIBRARY)
```
jserv: Don't append `$(LIBRABBITMQ_LIBRARY)` here. Instead, follow the consistent way to update `$(LDFLAGS).

docs/remote-dcurl.md
```diff
@@ -0,0 +1,34 @@
+# Remote dcurl
```
jserv: Rename "remote dcurl" to "remote interface" or "remote invocation".

remotedcurl/remotedcurl.c
```diff
@@ -0,0 +1,135 @@
...
+
+#include "amqp.h"
+#include "amqp_tcp_socket.h"
+
+#include "constants.h"
+#include "dcurl.h"
+
+#define HOSTNAME "localhost"
+#define PORT 5672
+#define HEATBEATS 0  // disables heartbeats
```
jserv: Why?

remotedcurl/remotedcurl.c
```diff
@@ -0,0 +1,135 @@
...
+#define ACCOUNT "guest"
```
jserv: Don't hardcode configurations here.

Makefile
```diff
@@ -158,6 +163,14 @@ $(OUT)/test-%: tests/test-%.py $(OUT)/libdcurl.so
 	$(call py_prepare_cmd)
 	$(Q)chmod +x $@
 
+$(OUT)/%.o: remotedcurl/%.c
```
jserv: Why did you create a new directory for `remotedcurl`? It can be consolidated with single source file.

marktwtn:
Since we will have our dcurl to be automatically released, the commit message should follow the specific format.

```
Type: Subject
```

You can see the different type from this [page](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#-commit-message-guidelines).

For example, you should write the commit like:
```
Feat: Implement remote dcurl with RabbitMQ broker 
Refactor: Remove unnecessary header files
```

Makefile
```diff
@@ -82,12 +82,17 @@ ifeq ("$(BUILD_COMPAT)", "1")
 TESTS += ccurl-multi_pow
 endif
 
+TESTS += remotedcurl-new-task
+TESTS += remotedcurl-get-result
```
marktwtn: Since the testing of remote interface is not as simple as other testing, I suggest to separate them.
Otherwise, the `make check` would never succeed.

I think we can classify them into two sets.
One is compile the testing and execute them directly, the other is compile but do not execute(which is suitable for remote interface).

docs/remote-dcurl.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Introduction
+
+Remote dcurl is inspired by [iotaledger's PoWbox](https://github.com/iotaledger/powbox). Remote dcurl supports asynchronous remote procedure call. We create two queues in RabbitMQ broker that manages queues and use [RabbitMQ C client](https://github.com/alanxz/rabbitmq-c) library to implement dcurl workers that do not compute accelerated PoW until users send new PoW tasks to the work queue (incomming queue) in RabbitMQ broker. In addition, dcurl worker-computed PoW result is sent to the queue (completed queue) in RabbitMQ broker and users can get PoW result from the queue. Users can see test-remotedcurl-new-task.c and test-remotedcurl-get-result.c to understand how to send a new PoW task and get a PoW result to/from work queues. By the way, It can surpport a scenario of a broker and many dcurl workers.
```
jserv: Illustrate the idea with ASCII-based diagrams. Also, you have to define these terminologies.

jserv:
Per discussion on Mar 4, we shall split the working items into several functional sub-tasks.

remotedcurl/remotedcurl.c
```diff
@@ -0,0 +1,135 @@
...
+#define INCOMMING_QUEUE "incomming_queue"
+#define COMPLETED_QUEUE "completed_queue"
+#define CHANNEL 1
+#define PASSIVE 0
+#define DURABLE 1
+#define EXCLUSIVE 0
+#define AUTODELETE 0
+#define EXCHANGE ""  // default exchange
+
+int main(int argc, char const *const *argv)
```
jserv: Function `main` is too long. Refactor it.

ajblane:
> Per discussion on Mar 4, we shall split the working items into several functional sub-tasks.

I try to write it down in issues #91

.gitmodules
ajblane: I only want to reserve a rabbitmq-c version for dcurl use. If we do not need it, I modify it.

tests/test-remotedcurl-get-result.c
```diff
@@ -0,0 +1,106 @@
+#include "common.h"
```
jserv: Rename to be consistent. Could be `test-remote-get.c`

tests/test-remotedcurl-new-task.c
```diff
@@ -0,0 +1,128 @@
+#include "common.h"
```
jserv: Ditto. Could be `test-remote-task.c`.

docs/remote-interface.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+Terminal 4: Get result of remote dcurl and verify PoW result
+```
+$./build/test-remotedcurl-get-result
+```
+
+## Limitations
+
+* Remote dcurl requires RabbitMQ broker
```
jserv, line 59: It is not limitation, but requirement.

ajblane:
> The feature of remote interface should be optional, that implies extra configuration(s).

The remote interface API is expected as the following link:
https://github.com/DLTcollab/dcurl/issues/91#issuecomment-469523999

remotedcurl/remotedcurl.c
ajblane: > Don't hardcode configurations here.

[process_all_options](https://github.com/alanxz/rabbitmq-c/blob/a65c64c0efd883f3e200bd8831ad3ca066ea523c/tools/common.c#L428) in RabbitMQ C client provides an implementation for options. Do we reuse it? Or, we choose an implementation for [simple .INI file parser](https://github.com/benhoyt/inih).

Makefile
jserv: Per the suggestion of @marktwtn, new test items should not be involved in RabbitMQ broker. You can append the test items which is decoupled with RabbitMQ.

remotedcurl/remotedcurl.c
jserv: `dcurl` is used in the form of library, and it would be suitable to configure it via default options and/or specific configuration file.

jserv:
Rebasing is required.

remotedcurl/remotedcurl.c
ajblane: We maybe require tow configuration files, one is for remote interface expected as the link https://github.com/DLTcollab/dcurl/issues/91#issuecomment-469523999 in client sides and the other is for dcurl workers.

And, Can I reuse [simple .INI file parser](https://github.com/benhoyt/inih) for configuration files?

ajblane: > `dcurl` is used in the form of library, and it would be suitable to configure it via default options and/or specific configuration file.

We maybe require two configuration files, one is for remote interface expected as the link https://github.com/DLTcollab/dcurl/issues/91#issuecomment-469523999 in client sides and the other is for dcurl workers.

And, Can I reuse [simple .INI file parser](https://github.com/benhoyt/inih) for configuration files?

jserv: Don't depend on configure file manipulation library at first glance. Think about the scenario of IRI + dcurl + remote interface where it is not suitable to specify extra files for configurations. Instead, we have to follow Java-style assignment. Taking the above into consideration, we shall eliminate the numbers of configuration entries as possible.

jserv: At the moment, only RMQ_HOST is required for configuration purpose. Something like `localhost:5672` might fit `RMQ_HOST`.

Makefile
```diff
@@ -82,17 +82,20 @@ ifeq ("$(BUILD_COMPAT)", "1")
 TESTS += ccurl-multi_pow
 endif
 
-TESTS += remote-new-task
-TESTS += remote-get-result
-
 TESTS := $(addprefix $(OUT)/test-, $(TESTS))
 
+MANUNAL_TESTS = \
```
jserv: Don't call it "manual tests", which sounds a bit sophisticated. Instead, we can use "trial", which implies the fashion of manual operations.

jserv:
Still conflicting with `develop` branch, fetch the latest commits and do `git rebase`.

docs/remote-interface.md
```diff
@@ -0,0 +1,59 @@
...
+ |                   Client                    |
+ |  +------------------+  +------------------+ |
+ |  |  send new work   |  |    get result    | |
+ |  +------------------+  +------------------+ |
+ |           |                     ^           |
+ +-----------|---------------------|-----------+
+ +-----------|---------------------|-----------+
+ |           v   RabbitMQ broker   |           |
+ |  +------------------+  +------------------+ |
+ |  | incommming queue |  | completed queue  | |
```
jserv: Fix typo.

docs/remote-interface.md
```diff
@@ -0,0 +1,59 @@
+# Remote interface
+
+## Introduction
+
+```
+ +---------------------------------------------+
+ |                   Client                    |
+ |  +------------------+  +------------------+ |
+ |  |  send new work   |  |    get result    | |
```
jserv, line 9: The essential operations should not contain only "send new work" and "get result". We need more descriptions.

docs/remote-interface.md
```diff
@@ -0,0 +1,59 @@
...
+             v                     |
+ +---------------------------------------------+
+ |                 dcurl worker                |
+ +---------------------------------------------+
+```
+
+Remote interface is inspired by [iotaledger's PoWbox](https://github.com/iotaledger/powbox). Remote interface supports asynchronous remote procedure call. We create two queues (incomming queue, completed queue) in RabbitMQ broker that can manages queues. Dcurl worker-required data is stored in the incomming queue by sending remote call of new work from client. After dcurl worker computes accelerated PoW, it sends remote call to store the PoW result in the completed queue in RabbitMQ broker. Moreover, we use [RabbitMQ C client](https://github.com/alanxz/rabbitmq-c) library to implement dcurl workers that do not compute accelerated PoW until users send data of new PoW tasks to the incomming queue in RabbitMQ broker. On the other hand, Users can see test-remote-new-task.c and test-remote-get-result.c to understand how to send a new PoW task and get a PoW result to/from work queues. By the way, It can surpport a scenario of a broker and many dcurl workers.
```
jserv: s/iotaledger's/IOTA's/

docs/remote-interface.md
```diff
@@ -0,0 +1,59 @@
+# Remote interface
+
+## Introduction
+
+```
```
jserv, line 5: ASCII-based diagram should be moved to the paragraph next to high-level descriptions.

mk/common.mk
```diff
@@ -37,5 +37,6 @@ $(OUT)/local.mk:
 
 clean:
 	$(RM) $(deps) $(TESTS) $(LIBS) $(OBJS)
+	$(RM) -r $(LIBRABBITMQ_PATH)/build
```
jserv, line 40: Don't mix RabbitMQ specific rules in `common.mk`. Instead, you can call predefined macros or execute target in RabbitMQ-oriented `.mk` file(s).

remotedcurl/remotedcurl.c
ajblane: In scenario of IRI + dcurl + remote interface, remote interface is composed of client, RabbitMQ broker and dcurl worker. As above discusses, we only requires RabbitMQ host IP setting and we use default values for other configurations in client side. On the other hand, dcurl worker also does this. In addition, for convenient setting of dcurl worker, I plan to use command line setting. 
```
./remotedcurl localhost 5672
```

jserv: The runtime parameter `localhost:5672` is expected to appear as an entry of IRI configurations. Keeping it simple and stupid is always a good policy.

remotedcurl/remotedcurl.c
ajblane: In my VM test environment, if heartbeat sets, the broker misses heartbeats from dcurl worker, default timeout is 1s. Error message is as follows: 
```
rabbitmq_1  | missed heartbeats from client, timeout: 1s
```

jserv: Check the manual of RabbitMQ  in explanation of specific configurations such as `HEATBEATS`.

Makefile
```diff
@@ -80,12 +81,24 @@ TESTS = \
 	dcurl \
 	pow
 
+ifeq ("$(BUILD_COMPAT)", "1")
```
jserv: Don't add merge-kludge as this.

Makefile
```diff
@@ -80,12 +81,24 @@ TESTS = \
...
+TESTS += ccurl-multi_pow
+endif
+
 TESTS := $(addprefix $(OUT)/test-, $(TESTS))
 
+TRIALS = \
```
jserv, line 86: Again, I dislike the operations you pre-define. Both `new-task` and `get-result` are not format terms we shall introduce. Be careful about the naming scheme.

Makefile
```diff
@@ -1,5 +1,6 @@
 OUT ?= ./build
 SRC := src
+SRC += -I remotedcurl
```
jserv: Is it necessary to create a new directory for remote interface? I expect single or two files added.

ajblane: I expect three files, remotedcurl.c -> remote_worker.c, remote_common.c and remote_interface.c. Yes, It's not necessary.

Makefile
```diff
@@ -12,7 +12,8 @@ ifneq ($(UNAME_S),Darwin)
 LDFLAGS += \
 	-Wl,--gc-sections \
 	-Wl,--as-needed \
-	-Wl,--version-script=./mk/libdcurl.version
+	-Wl,--version-script=./mk/libdcurl.version \
+	-Wl,--no-whole-archive
```
jserv, line 16: Why did you append `-Wl,--no-whole-archive`?

src/remote_worker.c
```diff
@@ -0,0 +1,90 @@
...
+    dcurl_init();
+
+    remote_state = connect_broker(&conn, CHANNEL);
+    if (remote_state == BROKER_CLOSE) {
+        printf("The rabbitmq broker is closed\n");
+        goto fail_to_exit;
+    } else if (remote_state == LOGIN_ERROR) {
+        printf("Login error \n");
+        goto fail_to_exit;
+    } else if (remote_state == LOGIN_ERROR) {
```
marktwtn, line 25: It should be `CHANNEL_ERROR`?

PR #118, opened 2019-03-03
Build: Extend build option for Sanitizer checking

marktwtn:
The extended and acceptable value of `BUILD_DEBUG` are
- 0: release build
- 1: debug build
- address: debug build + address Sanitizer
- undefined: debug build + undefined behavior Sanitizer
- thread: debug build + thread Sanitizer
If the unacceptable value is inputted, it would be debug build only.

The file mk/sanitizers.mk handles the GCC version validation
and filters out the invalid BUILD_DEBUG option.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,11 +17,14 @@ endif
 
 -include $(OUT)/local.mk
 
-ifeq ("$(BUILD_DEBUG)","1")
-CFLAGS += -Og -g3 -DENABLE_DEBUG
+ifneq ("$(BUILD_DEBUG)","0")
+    CFLAGS += -Og -g3 -DENABLE_DEBUG
+    ifneq ("$(BUILD_DEBUG)","1")
+        include mk/sanitizers.mk
+    endif
 else
-# Enable all the valid optimizations for standard programs in release build
-CFLAGS += -O3
+    # Enable all the valid optimizations for standard programs in release build
+    CFLAGS += -O3
 endif
 
 # Check specific CPU features available on build host
--- a/docs/build-n-test.md
+++ b/docs/build-n-test.md
@@ -22,6 +22,7 @@
     - ``BUILD_COMPAT``: build extra cCurl compatible interface.
     - ``BUILD_STAT``: show the statistics of the PoW information.
     - ``BUILD_DEBUG``: dump verbose messages internally.
+                       Build with the corresponding Sanitizer to detect software potential bugs if the value is `address`, `undefined` or `thread`.
 * Alternatively, you can specify conditional build as following:
 ```shell
 $ make BUILD_GPU=0 BUILD_JNI=1 BUILD_AVX=1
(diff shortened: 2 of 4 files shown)


docs/build-n-test.md
```diff
@@ -22,6 +22,7 @@
     - ``BUILD_COMPAT``: build extra cCurl compatible interface.
     - ``BUILD_STAT``: show the statistics of the PoW information.
     - ``BUILD_DEBUG``: dump verbose messages internally.
+    - ``BUILD_SAN``: build with Sanitizer to detect software potential bugs. Available values are `address`, `undefined` and `thread`.
```
jserv: How about migrating all relevant work for `BUILD_SAN` into `BUILD_DEBUG`? We can have both release and debug build.

marktwtn: Like remove the `BUILD_SAN` option and use `BUILD_DEBUG` option instead?
For example: The `BUILD_DEBUG` available option would be
- 0
- 1
- address
- undefined
- thread

jserv: I agree with the value range. The build with sanitizer(s) implies the set `CFLAGS += -Og -g3 -DENABLE_DEBUG`.

marktwtn: Here is what I'm going to do.
The `BUILD_SAN` option would be no longer existed.
Instead, the `BUILD_DEBUG` option will be used and the available option would extend to the corresponding Sanitizer.

The document will be modified as well.

jserv:
The subject is not reflecting the removal of Python-based test items.

marktwtn:
> The subject is not reflecting the removal of Python-based test items.

I guess I should separate the removal of Python testing code into another commit or
send another pull request instead?

jserv:
Go ahead for sending dedicated PR for Python removal.

marktwtn:
I update the subject and the first comment with the new commit message.
The **Python** related code would be handled in another pull request.

There is no more `BUILD_SAN` option at all.

docs/build-n-test.md
marktwtn: Done.

Makefile
```diff
@@ -17,11 +17,16 @@ endif
 
 -include $(OUT)/local.mk
 
-ifeq ("$(BUILD_DEBUG)","1")
-CFLAGS += -Og -g3 -DENABLE_DEBUG
+ifneq ("$(BUILD_DEBUG)","0")
+    CFLAGS += -Og -g3 -DENABLE_DEBUG
+    ifneq ("$(BUILD_DEBUG)","1")
+        # Sanitizer checking
```
jserv: We need extra `mk/sanitizers.mk` to filter out invalid input strings.

jserv: Also, gcc and/or clang versions should be validated in order to enable sanitizers.

marktwtn: Since not every Sanitizer is supported on `arm-linux`, I would focus on `x86` platform only.

marktwtn: Done in the new file `mk/sanitizers.mk`.

marktwtn:
The new file `mk/sanitizers.mk` is added for GCC version validation and filtering out the unacceptable input option.
The first comment of the pull request is updated with the new commit message.

PR #119, opened 2019-03-05
Test: Remove Python related code and build script

marktwtn:
The Python testing code has some problems with the GCC Sanitizers.
Since dcurl has already had C testing code for validating dcurl
implementation and the IOTA Python library is quite outdated,
the Python testing code can be removed safely.


wusyong:
Btw, maybe it's time to test if buildkite CI is working on pull request.

jserv:
Do we have any native tests equivalent to `multi-pow`?

marktwtn:
> Do we have any native tests equivalent to `multi-pow`?

No, we only have tests that execute a single PoW during the dcurl initialization and destruction.

Issue #95, opened 2019-01-31
Enable Sanitizers for CI and regular build targets


jserv:
@wusyong noticed potential memory misbehavior, and we should resolve these problems by enabling [sanitizers](https://github.com/google/sanitizers):
* AddressSanitizer: detects addressability issues;
* LeakSanitizer: detects memory leaks;
* ThreadSanitizer: detects data races and deadlocks
* MemorySanitizer: detects use of uninitialized memory

Expected output:
1. New build options to enable sanitizers;
2. Eliminate the issues pointed by sanitizers;
3. Travis-CI validates issues with sanitizers;

marktwtn:
From @jserv in https://github.com/DLTcollab/dcurl/issues/94#issuecomment-459228515

>I rebuilt `dcurl` with [AddressSanitizer](https://github.com/google/sanitizers/wiki/AddressSanitizer), and `make check` aborted with the following messages:
>```
>==52602==ERROR: LeakSanitizer: detected memory leaks
>...
>SUMMARY: AddressSanitizer: 37866102 byte(s) leaked in 663 allocation(s).
>```

marktwtn:
The `MemorySanitizer` can not be used since it is only supported in `LLVM` compiler,
as described in the **Getting MemorySanitizer** paragraph in the [Wiki(Msan)](https://github.com/google/sanitizers/wiki/MemorySanitizer) page.

And we do not have to use `LeakSanitizer` since it had been integrated into `AddressSanitizer` and enabled by default.
The [Wiki(LSan)](https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer) page also gives the warning:
```
Be aware that the stand-alone mode is less well tested
compared to running LSan on top of ASan.
```

marktwtn:
After integrating Sanitizers with flags `-fsanitize=[address|thread|undefined]`, it would generate error message when executing **Python** program with `AddressSanitizer`.

Error message:
```
==5001==ASan runtime does not come first in initial library list;
you should either link runtime to your application or manually preload it with LD_PRELOAD.
```

Although adding the environment variable `LD_PRELOAD` with `libasan` can remove the error message, but the memory leak detected by `AddressSanitizer` are mostly caused by **Python** itself.
In contrast with **Python**, the **C** source code has no problem at all, hence I decide to exclude the testing of `*.py`.

If there is no other advises, I will send a pull request tonight.

jserv:
We can discard Python related tests.

[...]

Issue #107, opened 2019-02-16
Unify thread synchronization APIs with libtuv


jserv:
Since `libtuv` is integrated into `dcurl`, we can unify the use of thread synchronization with [libtuv APIs](https://github.com/pando-project/libtuv/wiki/libtuv-APIs). Specifically, all occurrence of mutex and semaphore can be replaced with the corresponding libtuv equivalents, which are implemented with cross-platform support in mind. It implies that we can concentrate on abstract computing and resource management without enumerating GNU/Linux, macOS, and even MS-Windows ports.

jserv:
Related: #102

marktwtn:
Although replacing thread synchronization APIs with `libtuv` is trivial, it would have issue when compiling `dcurl` on other platform.
For example, the  `de10 nano` hardware platform.
Check #105 for reference.

The `libtuv` needs to specify the supported hardware and OS, but the current detection method in `mk/submodule.mk` is quite simple.

PR #128, opened 2019-03-21
fix: Avoid data race with read-write lock

marktwtn:
Use libtuv read-write lock API to avoid the data race
detected by thread Sanitizer.

Related #95.

### DIFF
--- a/src/pow_avx.c
+++ b/src/pow_avx.c
@@ -159,12 +159,15 @@ static int loop256(__m256i *lmid,
                    __m256i *hmid,
                    int m,
                    int8_t *nonce,
-                   int *stopPoW)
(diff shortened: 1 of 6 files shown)


wusyong (approved):
It seems work fine. Maybe we could add sanitizer step into each previous steps(generic, SSE...)

jserv:
Change the subject from "Fix data race" to "Avoid data race".

jserv:
We should not include `<pthread.h>` in each header file since we rely on `libtuv`.

marktwtn:
> 
> 
> We should not include `<pthread.h>` in each header file since we rely on `libtuv`.

Yes, I'll do it in another pull request for the issue #107.

PR #129, opened 2019-03-26
build: Allow libtuv to be built on other ARM boards

marktwtn:
The supported ARM boards are DE10-nano, Arrow SoCKit and Raspberry Pi 2/3 board.

Close #126.

PR #130, opened 2019-03-26
feat: Make thread synchronization portable

### LINKED ISSUE #107: Unify thread synchronization APIs with libtuv

marktwtn:
Replace the thread synchronization functions with the portable
libtuv APIs and remove the unused code.

TODO:
Make mutex of src/compat-ccurl.c portable.

Close #107.

### DIFF
--- a/src/dcurl.c
+++ b/src/dcurl.c
@@ -17,6 +17,7 @@
 #endif
 #include "implcontext.h"
 #include "trinary.h"
+#include "uv.h"
 #if defined(ENABLE_AVX)
 #include "pow_avx.h"
 #elif defined(ENABLE_SSE)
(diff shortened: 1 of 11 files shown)


wusyong (changes requested):
Block by #129

wusyong (approved):

Issue #92, opened 2019-01-28
Optimize Trits/Trytes manipulation with SIMD techniques


jserv:
Current tris/trytes manipulation in file `src/trinary.c` is quite naive, and it proceeds only a byte at one time:
```c
static bool validateTrytes(Trobject_t *trytes) {
    ...
    for (int i = 0; i < trytes->len; i++)
        if ((trytes->data[i] < 'A' || trytes->data[i] > 'Z') &&
            trytes->data[i] != '9')
            return false;
    return true;
}
```

[strcmp(3)](https://linux.die.net/man/3/strcmp) and its variants are crucial part of nearly every software project ever made. A modern CPU can process 32bit-256bit at a time inside a core, if it does only 8 bit in each operation then large parts of the computing capacity are unused. If the string code processes a lot of data and slows down the program it’s better to find some way to use this unused computing capability.

We can operate on each block rather than one byte to speed up string comparison as [Speeding up string compare and sorting](http://mgronhol.github.io/fast-strcmp/) suggested:
```c
int fast_compare(const char *ptr0, const char *ptr1, int len) {
    int fast = len / sizeof(size_t) + 1;
    int offset = (fast - 1) * sizeof(size_t);
    int current_block = 0;
    if (len <= sizeof(size_t)) { fast = 0; }
    size_t *lptr0 = (size_t*) ptr0;
    size_t *lptr1 = (size_t *) ptr1;

    while (current_block < fast) {
        if ((lptr0[current_block] ^ lptr1[current_block])) {
            for (int pos = current_block * sizeof(size_t); pos < len ; ++pos) {
                if ((ptr0[pos] ^ ptr1[pos]) || (ptr0[pos] == 0) || (ptr1[pos] == 0))
                    return  (int)((unsigned char) ptr0[pos] - (unsigned char) ptr1[pos]);
            }
        }
        ++current_block;
    }

    while (len > offset) {
        if ((ptr0[offset] ^ ptr1[offset]))
            return (int)((unsigned char) ptr0[offset] - (unsigned char) ptr1[offset]); 
        ++offset;
    }	
    return 0;
}
```

[...]

PR #135, opened 2019-04-04
perf: Optimize trytes from trits with x86 SIMD

marktwtn:
Without SIMD optimization:
Input size(byte) - Average time(nsec)
81               - 406.3
243              - 444.6

With SIMD optimization:
Input size(byte) - Average time(nsec)
81               - 261.5
243              - 162.7

Hardware information:
architecture - x86_64
CPU          - AMD Ryzen 5 2400G

Related #92.

### DIFF
--- a/src/trinary.c
+++ b/src/trinary.c
@@ -35,12 +35,11 @@ static bool validateTrits(Trobject_t *trits)
 
 #if defined(__SSE4_2__)
     return validateTrits_sse42(trits);
-#else
+#endif
     for (int i = 0; i < trits->len; i++)
         if (trits->data[i] < -1 || trits->data[i] > 1)
             return false;
     return true;
-#endif
 }
 
 static bool validateTrytes(Trobject_t *trytes)
@@ -50,13 +49,12 @@ static bool validateTrytes(Trobject_t *trytes)
 
 #if defined(__SSE4_2__)
     return validateTrytes_sse42(trytes);
-#else
+#endif
     for (int i = 0; i < trytes->len; i++)
         if ((trytes->data[i] < 'A' || trytes->data[i] > 'Z') &&
             trytes->data[i] != '9')
             return false;
     return true;
-#endif
 }
 
 Trobject_t *initTrits(int8_t *src, int len)
@@ -135,7 +133,7 @@ Trobject_t *trytes_from_trits(Trobject_t *trits)
 
 #if defined(__SSE4_2__)
     return trytes_from_trits_sse42(trits);
-#else
+#endif
     Trobject_t *trytes = NULL;
     int8_t *src = (int8_t *) malloc(trits->len / 3);
 
@@ -153,7 +151,6 @@ Trobject_t *trytes_from_trits(Trobject_t *trits)
     free(src);
 
     return trytes;
-#endif
 }
 
 Trobject_t *trits_from_trytes(Trobject_t *trytes)
--- a/src/trinary.c
+++ b/src/trinary.c
@@ -133,6 +133,9 @@ Trobject_t *trytes_from_trits(Trobject_t *trits)
         return NULL;
     }
 
+#if defined(__SSE4_2__)
+    return trytes_from_trits_sse42(trits);
+#else
     Trobject_t *trytes = NULL;
     int8_t *src = (int8_t *) malloc(trits->len / 3);
 
@@ -150,6 +153,7 @@ Trobject_t *trytes_from_trits(Trobject_t *trits)
     free(src);
 
     return trytes;
+#endif
 }
 
 Trobject_t *trits_from_trytes(Trobject_t *trytes)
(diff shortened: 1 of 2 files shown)


src/trinary_sse42.h
```diff
@@ -72,4 +73,161 @@ static inline bool validateTrytes_sse42(Trobject_t *trytes)
...
 
+static inline Trobject_t *trytes_from_trits_sse42(Trobject_t *trits)
+{
+    Trobject_t *trytes = NULL;
+    int8_t *src = (int8_t *) malloc(trits->len / 3);
+
+    const int block_8bit = BLOCK_8BIT(__m128i);
+    int8_t setMSB = 0x80;
+    __m128i tryteAlphabet[2] = {
+        _mm_setr_epi8('9', 'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J',
```
jserv: Why did you enumerate the alphabet between 'A' to 'Z'?

marktwtn: For the later use of transforming the trit to tryte.
It is similar to the `TryteAlphabet` defined in `src/constants.c`.

jserv: Is it possible to introduce macros for such enumeration?

src/trinary_sse42.h
```diff
@@ -72,4 +73,161 @@ static inline bool validateTrytes_sse42(Trobject_t *trytes)
...
+                      'K', 'L', 'M', 'N', 'O'),
+        _mm_setr_epi8('P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z', 0,
+                      0, 0, 0, 0)};
+    /* For shuffling the bytes of the input trits */
```
jserv, line 113: I wish there are some macros for shuffling, otherwise it is almost impossible to maintain.

src/trinary_sse42.h
marktwtn: Can I have more clues that the code modification should probably look like?

jserv: https://github.com/jserv/sse2neon/blob/master/sse2neon.h # Check macro `_MM_SHUFFLE`

marktwtn: I only substituted them with the variable `TryteAlphabet`.

src/trinary_sse42.h
marktwtn: I still have no idea about the better way to introduce macros for shuffling.
So I make the code remain the same.

Here comes two ways to introduce macros:
1. With many operations of shifting, masking and setting value.
However, it would be too tedious to give value for a `__m128i` array.
The performance would be affected, too.
2. Another way with macros:
```
     const __m128i shuffleLow[3] = {
-        _mm_setr_epi8(0x00, 0x03, 0x06, 0x09, 0x0C, 0x0F, setMSB, setMSB,
-                      setMSB, setMSB, setMSB, setMSB, setMSB, setMSB, setMSB,
-                      setMSB),
-        _mm_setr_epi8(setMSB, setMSB, setMSB, setMSB, setMSB, setMSB, 0x02,
-                      0x05, 0x08, 0x0B, 0x0E, setMSB, setMSB, setMSB, setMSB,
-                      setMSB),
-        _mm_setr_epi8(setMSB, setMSB, setMSB, setMSB, setMSB, setMSB, setMSB,
-                      setMSB, setMSB, setMSB, setMSB, 0x01, 0x04, 0x07, 0x0A,
-                      0x0D)};
+        _mm_setr_epi8(REPEAT(0, setMSB) COMMA(0) INDEX_3DIFF_0F COMMA(1)
+                          REPEAT(10, setMSB)),
+        _mm_setr_epi8(REPEAT(6, setMSB) COMMA(1) INDEX_3DIFF_2E COMMA(1)
+                          REPEAT(5, setMSB)),
+        _mm_setr_epi8(REPEAT(11, setMSB) COMMA(1) INDEX_3DIFF_1D COMMA(0)
+                          REPEAT(0, setMSB))};

```
I am not sure that this is acceptable or not.
It seems not helping much but it looks more elegant.

jserv: The later looks better to me.

src/trinary.c
```diff
@@ -133,6 +133,9 @@ Trobject_t *trytes_from_trits(Trobject_t *trits)
         return NULL;
     }
 
+#if defined(__SSE4_2__)
+    return trytes_from_trits_sse42(trits);
+#else
```
jserv: We can safely replace `#else` with `#endif` since it exists as a shortcut.

marktwtn: Done with the additional commit `refactor: Remove redundant conditional compilation`.

src/trinary_sse42.h
marktwtn: They are all replaced with the latter format.

PR #141, opened 2019-04-24
Jar create and extract

### LINKED ISSUE #106: Provide Java package for deployment
Current JNI build process generates shared library in the form of pearldiver driver for IRI. However, it is not convenient for deployment as RocksDB does. By providing `rocksdbjni` package, maven can find and package `rocksdbjni` into final IRI jar file including RocksDB C++ library.

Once we can provide similar java package (likely `dcurl-jni`), we can simplify the process of validation and deployment.

marktwtn:
- feat: Create JAR file for Java native interface
Build with option `BUILD_JNI=1` would compile the Java code and create the JAR file.
The dcurl shared library `libdcurl.so` would be packaged into the JAR file
and can be extracted from it for Jave native interface easily.
Close #106.

- docs: Change the information of JNI and IRI

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+VERSION = 0.2.0
+
 OUT ?= ./build
 SRC := src
 
@@ -89,7 +91,15 @@ TESTS := $(addprefix $(OUT)/test-, $(TESTS))
 LIBS = libdcurl.so
 LIBS := $(addprefix $(OUT)/, $(LIBS))
 
-all: config $(TESTS) $(LIBS)
+JARS := dcurljni-$(VERSION).jar
+JARS := $(addprefix $(OUT)/, $(JARS))
+
+PREQ := config $(TESTS) $(LIBS)
+ifeq ("$(BUILD_JNI)","1")
+PREQ += $(JARS)
+endif
+
+all: $(PREQ)
 .DEFAULT_GOAL := all
 
 OBJS = \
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -1,3 +1,5 @@
+VERSION = 0.2.0
```
jserv, line 1: Should the `VERSION` be specific to ABI or synchronized with each release?

README.md
```diff
@@ -28,7 +28,13 @@ After integrating dcurl into IRI, performance of [attachToTangle](https://iota.r
 
 ## IRI Adaptation
 [Modified IRI accepting external PoW Library](https://github.com/DLTcollab/iri)
-Supported IRI version: 1.6.0
+Supported IRI version: 1.7.0
```
jserv, line 31: Can we provide backward JNI support?

java/org/dcurl/Dcurl.java
```diff
@@ -0,0 +1,43 @@
+package org.dcurl;
```
jserv: The package name should be `org.dltcollab`.

README.md
marktwtn: Like support the old way to copy the `libdcurl.so` file and assign the correct path?
If so, it is still supported in IRI.
I just mark it as deprecated in the `README.md` since I guess it might be replaced by the new loading method.

Makefile
marktwtn: In my original design, it should be synchronized with each release.

java/org/dcurl/Dcurl.java
marktwtn: It has been renamed.

PR #142, opened 2019-04-25
perf: Increase the default thread pool size of libtuv

marktwtn:
The dcurl can not affect the thread pool size of libtuv.
For allowing more threads to calculate the PoW,
the size must be increased.
The default pool size is changed from 4 to 26.


jserv:
I dislike such modifications. The default value is valid for libtuv's motivation and purpose on IoT devices. We can refer to runtime configurations in order to calibrate CPU resources. Also, consider the case when HyperThreading is enabled.

marktwtn:
I'll figure out another method for changing the thread pool size.
The pull request is not going to be merged.

PR #143, opened 2019-04-25
feat: Remote interface interoperate distributed computing resources

### LINKED ISSUE #137: Remote interface: Implement client side for RPC
Currently,  RabbitMQ C client only provides [the RPC client example](https://github.com/alanxz/rabbitmq-c/blob/1eb5fb464561bf44e571c4d57a5a63cfe3ca5c33/examples/amqp_rpc_sendstring_client.c) and not provides the RPC server example. 
## RPC server
Here are some samples:
- [x]  In the issue, ["Add rpc_server example #394"](https://github.com/alanxz/rabbitmq-c/issues/394), there is a [rpc-server implementation](https://github.com/alanxz/rabbitmq-c/compare/master...wallkietallkie:patch-1) that only can run with some problems that are a invalid pointer when freeing and no acknowledge. We can fix these problems and implement dcurl workers based on this implementation. 
* [SimpleRpcServer](https://github.com/kkirtac/SimpleRpcServer)

## RPC client 
- [x]  Exclusive callback queues with TTL property
For the RPC client example, it declares a private queue (callback queue) with auto-delete that means the queue is deleted when all consumers have finished using it. However, one RPC client only has one private queue. Therefore, I will modify a private queue with auto-delete into a private queue with exclusive. See [RabbitMQ the difference between exclusive and auto-delete](https://stackoverflow.com/questions/21248563/rabbitmq-difference-between-exclusive-and-auto-delete?fbclid=IwAR1vWMwTOFvQU5tdpxnYo4sxWkjK-B3148OJrFAGOQkhcxaaFHuKorqMHEw) to know them. Moreover, exclusive queues auto-delete when the consumer disconnects and we need to keep connection alive, so we need to use exclusive callback queues with [TTL property](https://www.rabbitmq.com/ttl.html) to automatically delete callback queues.
- [x]  Correlation id is not required
In addition, in [ the scenario of AttachToTangle API](https://github.com/DLTcollab/dcurl/issues/133), there only is 0 or 1 result message in the callback queue of a AttachToTangle thread, so correlation id is not required. 
- [x] ~Channel management for multiple threads~
It is very common to open a new channel per thread and not share channels between them, see ["Channels"](https://www.rabbitmq.com/channels.html), so we may require the channel management. Here is [a example of the implementation of channel management].(https://github.com/alanxz/SimpleAmqpClient/blob/master/src/ChannelImpl.cpp#L134)
- [x] AMQP connection management for multiple threads
For the alanxz-provided RabbitMQ C AMQP client library, it does not support that multiple threads execute multiple channels in an AMQP connection. Therefore, we instead should open an AMQP connection per thread, as refers to ["Writing applications using librabbitmq"](https://github.com/alanxz/rabbitmq-c#threading).

ajblane:
* make BUILD_REMOTE=1 to build libdcurl.so and remote-worker
* make BUILD_REMOTE=1 check with RabbitMQ broker and remote-worker
* RPC with exclusive callback queues with TTL property
* AMQP connection management for multiple threads
* Implement local fallback PoW when remote interface fails

Resolved #137

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,10 @@ ifeq ("$(BUILD_FPGA_ACCEL)","1")
 CFLAGS += -DENABLE_FPGA_ACCEL
 endif
 
+ifeq ("$(BUILD_REMOTE)","1")
+CFLAGS += -DENABLE_REMOTE
+endif
+
 ifeq ("$(BUILD_JNI)","1")
 include mk/java.mk
 endif
@@ -98,6 +102,9 @@ JARS := $(addprefix $(OUT)/, $(JARS))
 PREQ := config $(TESTS) $(LIBS)
 ifeq ("$(BUILD_JNI)","1")
 PREQ += $(JARS)
+endif
+ifeq ("$(BUILD_REMOTE)", "1")
+PREQ += $(OUT)/remote-worker
 endif
 
 all: $(PREQ)
@@ -140,23 +147,36 @@ OBJS += \
 	pow_fpga_accel.o
 endif
 
+ifeq ("$(BUILD_REMOTE)", "1")
+OBJS += \
+	remote_common.o \
+	remote_interface.o
+
+WORKER_OBJS := $(addprefix $(OUT)/worker-,$(filter-out remote_interface.o, $(OBJS)))
+WORKER_CFLAGS := $(filter-out -DENABLE_REMOTE, $(CFLAGS))
+endif
+
 OBJS := $(addprefix $(OUT)/, $(OBJS))
 
 $(OUT)/test-%.o: tests/test-%.c $(LIBTUV_PATH)/include
 	$(VECHO) "  CC\t$@\n"
 	$(Q)$(CC) -o $@ $(CFLAGS) -I $(SRC) $(LIBTUV_INCLUDE) -c -MMD -MF $@.d $<
 
-$(OUT)/%.o: $(SRC)/%.c $(LIBTUV_PATH)/include
+$(OUT)/%.o: $(SRC)/%.c $(LIBTUV_PATH)/include $(LIBRABBITMQ_PATH)/build/include
 	$(VECHO) "  CC\t$@\n"
-	$(Q)$(CC) -o $@ $(CFLAGS) $(LIBTUV_INCLUDE) -c -MMD -MF $@.d $<
+	$(Q)$(CC) -o $@ $(CFLAGS) $(LIBTUV_INCLUDE) $(LIBRABBITMQ_INCLUDE) -c -MMD -MF $@.d $<
 
-$(OUT)/test-%: $(OUT)/test-%.o $(OBJS) $(LIBTUV_LIBRARY)
+$(OUT)/test-%: $(OUT)/test-%.o $(OBJS) $(LIBTUV_LIBRARY) $(LIBRABBITMQ_LIBRARY)
 	$(VECHO) "  LD\t$@\n"
-	$(Q)$(CC) -o $@ $^ $(LDFLAGS)
+	$(Q)$(CC) -o $@ $^ $(LDFLAGS) $(LIBRABBITMQ_LINK)
 
-$(OUT)/libdcurl.so: $(OBJS) $(LIBTUV_LIBRARY)
+$(OUT)/libdcurl.so: $(OBJS) $(LIBTUV_LIBRARY) $(LIBRABBITMQ_LIBRARY)
 	$(VECHO) "  LD\t$@\n"
-	$(Q)$(CC) -shared -o $@ $^ $(LDFLAGS)
+	$(Q)$(CC) -shared -o $@ $^ $(LDFLAGS) $(LIBRABBITMQ_LINK)
+
+ifeq ("$(BUILD_REMOTE)", "1")
+include mk/remote.mk
+endif
 
 include mk/common.mk
 
--- a/docs/remote-interface.md
+++ b/docs/remote-interface.md
@@ -1,0 +1,67 @@
+# Remote interface
+## Introduction
+
+```
+ +-----------------------------------------------+
+ |               remote interface                |
+ |    +----------------------------------------+ | 
+ |   +----------------------------------------+| |
+ |  +----------------------------------------+|+ |
+ |  |     RabbitMQ-provided RPC              |+  |
+ |  +----------------------------------------+   |
+ |            |                     ^            |
+ +------------|---------------------|------------+
+ +------------|---------------------|------------+
+ |            |   RabbitMQ broker   |            |
+ |            |                     |            |
+ |            |             +------------------+ |
+ |            v            +------------------+| |
+ |  +------------------+  +------------------+|| |
+ |  | incoming queue   |  | private queue    ||| |
+ |  |                  |  |                  ||| |
+ |  | - trytes         |  | - PoW result     ||| |
+ |  | - mwm            |  |                  ||+ |
+ |  |                  |  |                  |+  |
+ |  +------------------+  +------------------+   |
+ |            |                     ^            |
+ +------------|---------------------|------------+
+              v                     |
+   +---------------------------------------------+
+  +---------------------------------------------+|
+ +---------------------------------------------+|+
+ |                  remote worker              |+
+ +---------------------------------------------+
+```
+To support asynchronous remote procedure call, remote interface in dcurl provides an interface named as `Remote_ImplContext` to implement it. dcurl currently uses RabbitMQ C client to implement asynchronous RPC in remote interface. Remote interface provides thread management to support an asynchronous RPC per thread.
+
+Here are detailed implementations of the RabbitMQ-provided RPC pattern as follows:
+* Asynchronous RPC requests are inserted into the message queue, `incoming_queue`, in RabbitMQ broker
+* Asynchronous RPCs with exclusive private queues (callback queues) with TTL = 10s property
+* Correlation ID is not used
+* An asynchronous RPC uses a connection to RabbitMQ broker
+* Remote workers can obtain requests from `incoming_queue` by default exchange of RabbitMQ broker
+
+## How to test remote interface in localhost
+You need to open three terminals
+
+Terminal 1: Run the RabbitMQ broker You can quickly use docker to run the RabbitMQ broker, rabbitmq
+```
+$ sudo docker run -d rabbitmq
+```
+
+Terminal 2: Run remote workers
+```
+$ ./build/remote-worker
+```
+How to build remote worker on FPGA board
+```
+$ make BUILD_REMOTE=1 BUILD_FPGA_ACCEL=1 BOARD=de10nano
+```
+
+Terminal 3: Run check
+```
+$ make BUILD_REMOTE=1 BUILD_DEBUG=1 check
+```
+
+## Requirements
+Remote interface requires RabbitMQ broker
--- a/mk/defs.mk
+++ b/mk/defs.mk
@@ -19,6 +19,9 @@ BUILD_GPU ?= 0
 # Build FPGA backend or not
 BUILD_FPGA_ACCEL ?= 0
 
+# Build facilities of remote procedure calls
+BUILD_REMOTE ?= 0
+
 # Build JNI glue as the bridge between dcurl and IRI
 BUILD_JNI ?= 0
 
--- a/mk/remote.mk
+++ b/mk/remote.mk
@@ -1,0 +1,8 @@
+# Build remote-worker
+$(OUT)/worker-%.o: $(SRC)/%.c $(LIBTUV_PATH)/include $(LIBRABBITMQ_PATH)/build/include
+	$(VECHO) "  CC\t$@\n"
+	$(Q)$(CC) -o $@ $(WORKER_CFLAGS) $(LIBTUV_INCLUDE) $(LIBRABBITMQ_INCLUDE) -c -MMD -MF $@.d $<
+
+$(OUT)/remote-worker: $(OUT)/remote_worker.o $(WORKER_OBJS) $(LIBTUV_LIBRARY) $(LIBRABBITMQ_LIBRARY)
+	$(VECHO) "  LD\t$@\n"
+	$(Q)$(CC) -o $@ $^ $(LDFLAGS) $(LIBRABBITMQ_LINK)
--- a/mk/submodule.mk
+++ b/mk/submodule.mk
@@ -26,3 +26,34 @@ $(LIBTUV_PATH)/include:
 
 $(LIBTUV_LIBRARY):
 	$(MAKE) -C $(LIBTUV_PATH) TUV_BUILD_TYPE=release TUV_CREATE_PIC_LIB=yes TUV_PLATFORM=$(LIBTUV_PLATFORM) TUV_BOARD=$(LIBTUV_BOARD)
+
+# librabbitmq related variables
+LIBRABBITMQ_PATH = deps/rabbitmq-c
+LIBRABBITMQ_INCLUDE := -I $(LIBRABBITMQ_PATH)/build/include
+LIBRABBITMQ_LIB_PATH := $(LIBRABBITMQ_PATH)/build/librabbitmq/
+ifeq ($(UNAME_S),darwin)
+    # macOS
+    LIBRABBITMQ_LINK := -Wl,-rpath,$(LIBRABBITMQ_LIB_PATH) -L$(LIBRABBITMQ_LIB_PATH) -lrabbitmq
+    LIBRABBITMQ_LIBRARY := $(LIBRABBITMQ_LIB_PATH)/librabbitmq.dylib
+else
+    LIBRABBITMQ_LINK := -Wl,-rpath=$(LIBRABBITMQ_LIB_PATH) -L$(LIBRABBITMQ_LIB_PATH) -lrabbitmq
+    LIBRABBITMQ_LIBRARY := $(LIBRABBITMQ_LIB_PATH)/librabbitmq.so
+endif
+
+$(LIBRABBITMQ_PATH)/build/include:
+	git submodule update --init $(LIBRABBITMQ_PATH)
+	mkdir $(LIBRABBITMQ_PATH)/build
+ifeq ($(UNAME_S),darwin)
+	# macOS
+	cd $(LIBRABBITMQ_PATH)/build && \
+         cmake -DOPENSSL_ROOT_DIR=/usr/local/opt/openssl/ -DCMAKE_INSTALL_PREFIX=. .. && \
+         cmake --build . --target install
+else
+	cd $(LIBRABBITMQ_PATH)/build && \
+         cmake -DCMAKE_INSTALL_PREFIX=. .. && \
+         cmake --build . --target install
+endif
+
+$(LIBRABBITMQ_LIBRARY):
+	cd $(LIBRABBITMQ_PATH)/build && \
+         cmake --build .
--- a/src/dcurl.c
+++ b/src/dcurl.c
@@ -15,6 +15,9 @@
 #if defined(ENABLE_FPGA_ACCEL)
 #include "pow_fpga_accel.h"
 #endif
+#if defined(ENABLE_REMOTE)
+#include "remote_interface.h"
+#endif
 #include "implcontext.h"
 #include "trinary.h"
 #include "uv.h"
@@ -48,6 +51,11 @@ extern ImplContext PoWCL_Context;
 extern ImplContext PoWFPGAAccel_Context;
 #endif
 
+#if defined(ENABLE_REMOTE)
+extern RemoteImplContext Remote_Context;
+static uv_sem_t notify_remote;
+#endif
+
 bool dcurl_init()
 {
     bool ret = true;
@@ -68,6 +76,11 @@ bool dcurl_init()
     ret &= registerImplContext(&PoWFPGAAccel_Context);
 #endif
 
+#if defined(ENABLE_REMOTE)
+    ret &= initializeRemoteContext(&Remote_Context);
+    uv_sem_init(&notify_remote, 0);
+#endif
+
     uv_sem_init(&notify, 0);
     return isInitialized = ret;
 }
@@ -76,6 +89,10 @@ void dcurl_destroy()
 {
     ImplContext *impl = NULL;
     struct list_head *p;
+
+#if defined(ENABLE_REMOTE)
+    destroyRemoteContext(&Remote_Context);
+#endif
 
     list_for_each (p, &IMPL_LIST) {
         impl = list_entry(p, ImplContext, list);
@@ -96,6 +113,28 @@ int8_t *dcurl_entry(int8_t *trytes, int mwm, int threads)
     if (!isInitialized)
         return NULL;
 
+#if defined(ENABLE_REMOTE)
+    do {
+        if (enterRemoteContext(&Remote_Context)) {
+            pow_ctx = getRemoteContext(&Remote_Context, trytes, mwm);
+            goto remote_pow;
+        }
+        uv_sem_wait(&notify_remote);
+    } while (1);
+
+remote_pow:
+    if (!doRemoteContext(&Remote_Context, pow_ctx)) {
+        goto local_pow;
+    } else {
+        res = getRemoteResult(&Remote_Context, pow_ctx);
+    }
+    freeRemoteContext(&Remote_Context, pow_ctx);
+    exitRemoteContext(&Remote_Context);
+    uv_sem_post(&notify_remote);
+    return res;
+
+local_pow:
+#endif
     do {
         list_for_each (p, &IMPL_LIST) {
             impl = list_entry(p, ImplContext, list);
--- a/src/remote_common.c
+++ b/src/remote_common.c
@@ -1,0 +1,323 @@
+/*
+ * Copyright (C) 2019 dcurl Developers.
+ * Use of this source code is governed by MIT license that can be
+ * found in the LICENSE file.
+ */
+
+#include "remote_common.h"
+#include <stdio.h>
+#include <string.h>
+#include "common.h"
+
+bool die_on_amqp_error(amqp_rpc_reply_t x, char const *context)
+{
+    switch (x.reply_type) {
+    case AMQP_RESPONSE_NORMAL:
+        return true;
+
+    case AMQP_RESPONSE_NONE:
+        ddprintf("%s: missing RPC reply type!\n", context);
+        break;
+
+    case AMQP_RESPONSE_LIBRARY_EXCEPTION:
+        ddprintf("%s: %s\n", context,
+                amqp_error_string2(x.library_error));
+        break;
+
+    case AMQP_RESPONSE_SERVER_EXCEPTION:
+        switch (x.reply.id) {
+        case AMQP_CONNECTION_CLOSE_METHOD: {
+            amqp_connection_close_t *m =
+                (amqp_connection_close_t *) x.reply.decoded;
+            ddprintf("%s: server connection error %uh, message: %.*s\n",
+                    context, m->reply_code, (int) m->reply_text.len,
+                    (char *) m->reply_text.bytes);
+            break;
+        }
+        case AMQP_CHANNEL_CLOSE_METHOD: {
+            amqp_channel_close_t *m = (amqp_channel_close_t *) x.reply.decoded;
+            ddprintf("%s: server channel error %uh, message: %.*s\n",
+                    context, m->reply_code, (int) m->reply_text.len,
+                    (char *) m->reply_text.bytes);
+            break;
+        }
+        default:
+            ddprintf("%s: unknown server error, method id 0x%08X\n",
+                    context, x.reply.id);
+            break;
+        }
+        break;
+    }
+
+    return false;
+}
+
+bool die_on_error(int x, char const *context)
+{
+    if (x < 0) {
+        ddprintf("%s: %s\n", context, amqp_error_string2(x));
+        return false;
+    }
+
+    return true;
+}
+
+bool connect_broker(amqp_connection_state_t *conn)
+{
+    amqp_socket_t *socket = NULL;
+
+    /* Connect to the rabbitmq broker */
+    *conn = amqp_new_connection();
+    socket = amqp_tcp_socket_new(*conn);
+    if (amqp_socket_open(socket, HOSTNAME, 5672) != AMQP_STATUS_OK) {
+        ddprintf("%s\n", "The rabbitmq broker is closed");
+        goto destroy_connection;
+    }
+
+    /* Login to the rabbitmq broker */
+    if (!die_on_amqp_error(amqp_login(*conn, "/", AMQP_DEFAULT_MAX_CHANNELS,
+                                      AMQP_DEFAULT_FRAME_SIZE, 0,
+                                      AMQP_SASL_METHOD_PLAIN, "guest", "guest"),
+                           "Logging in"))
+        goto connection_close;
+
+    /* Open the channel in the rabbitmq broker */
+    amqp_channel_open(*conn, 1);
+    if (!(die_on_amqp_error(amqp_get_rpc_reply(*conn),
+                                "Opennng the channel")))
+        goto channel_close;
+
+    return true;
+
+channel_close:
+    amqp_channel_close(*conn, 1, AMQP_REPLY_SUCCESS);
+
+connection_close:
+    amqp_connection_close(*conn, AMQP_REPLY_SUCCESS);
+
+destroy_connection:
+    amqp_destroy_connection(*conn);
+
+    return false;
+}
+
+bool declare_queue(amqp_connection_state_t *conn,
+                   amqp_channel_t channel,
+                   char const *queue_name)
+{
+    /* Declare a durable queue */
+    amqp_queue_declare(*conn, channel, amqp_cstring_bytes(queue_name), 0, 1, 0,
+                       0, amqp_empty_table);
+    if (!die_on_amqp_error(amqp_get_rpc_reply(*conn), "Declaring the queue"))
+        return false;
+
+    return true;
+}
+
+bool declare_callback_queue(amqp_connection_state_t *conn,
+                            amqp_channel_t channel,
+                            amqp_bytes_t *reply_to_queue)
+{
+    /* Declare a exclusive private queues with TTL = 10s */
+    amqp_table_entry_t entries[1];
+    amqp_table_t table;
+    entries[0].key = amqp_cstring_bytes("x-message-ttl");
+    entries[0].value.kind = AMQP_FIELD_KIND_U32;
+    entries[0].value.value.u32 = 10 * 1000;  // 10s
+    table.num_entries = 1;
+    table.entries = entries;
+
+    amqp_queue_declare_ok_t *r =
+        amqp_queue_declare(*conn, channel, amqp_empty_bytes, 0, 0, 1, 0, table);
+    if (!die_on_amqp_error(amqp_get_rpc_reply(*conn),
+                           "Declaring the private queue with TTL = 10s"))
+        return false;
+
+    *reply_to_queue = amqp_bytes_malloc_dup(r->queue);
+    return true;
+}
+
+bool set_consuming_queue(amqp_connection_state_t *conn,
+                         amqp_channel_t channel,
+                         char const *queue_name)
+{
+    amqp_basic_consume(*conn, channel, amqp_cstring_bytes(queue_name),
+                       amqp_empty_bytes, 0, 0, 0, amqp_empty_table);
+    if (!die_on_amqp_error(amqp_get_rpc_reply(*conn),
+                           "Set the consuming queue"))
+        return false;
+
+    return true;
+}
+
+bool consume_message(amqp_connection_state_t *conn,
+                     amqp_channel_t channel,
+                     amqp_envelope_t *envelope)
+{
+    amqp_maybe_release_buffers(*conn);
+    /* Consume a message from the queue in the rabbitmq broker */
+    if (!die_on_amqp_error(amqp_consume_message(*conn, envelope, NULL, 0),
+                           "Consuming the message"))
+        return false;
+
+    return true;
+}
+
+bool wait_response_message(amqp_connection_state_t *conn,
+                           amqp_channel_t channel,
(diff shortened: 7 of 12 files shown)


wusyong:
Commit should be atomic. Please consider rebase to several ones. Also some test cases would be nice.

src/remote_worker.c
```diff
@@ -0,0 +1,78 @@
...
+        goto fail_to_exit;
+
+    if (!set_consuming_queue(&conn, CHANNEL, "incoming_queue"))
+        goto fail_to_exit;
+
+    for (;;) {
+        if (!consume_message(&conn, CHANNEL, &envelope))
+            goto fail_to_exit;
+
+        printf(
```
jserv: Don't invoke `printf` directly. Instead, ensure conditional build and optional diagnosis mode for message display.

src/remote_worker.c
```diff
@@ -0,0 +1,78 @@
...
+        /* Publish a message of remote PoW result */
+        if (!publish_message(
+                &conn, CHANNEL,
+                (char *) envelope.message.properties.reply_to.bytes,
+                (char *) ret_trytes))
+            goto fail_to_exit;
+
+        free(ret_trytes);
+        amqp_destroy_envelope(&envelope);
+        printf("Publishing PoW result to callback queue is done\n");
```
jserv: Ditto. Eliminate the need of manipulating stdout.

src/remote_worker.c
```diff
@@ -0,0 +1,78 @@
+/*
+ * Copyright (C) 2018 dcurl Developers.
```
jserv: Should be `2019`.

src/remote_interface.c
```diff
@@ -0,0 +1,262 @@
...
+        }
+    }
+
+    uv_mutex_unlock(&remoteinterface_ctx->lock);
+
+    return NULL; /* It should not happen */
+}
+
+static bool PoWRemoteInterface_freePoWContext(
+    RemoteInterfaceContext *remoteinterface_ctx,
```
jserv: `remoteinterface_ctx` might not a good name. Instead, simply using `ctx` should work out of box.

ajblane: In the remote_interface.c, `ctx` name has been used.  Therefore, for simplicity, `remoteinterface_ctx ` will be renamed into `remote_ctx`.

jserv: By the way, the naming `PoWRemoteInterface_freePoWContext` is really redundant. It was intended to be a helper function . Simply calling it as `RI_freePoWcontext` works well.

Makefile
```diff
@@ -139,23 +149,56 @@ OBJS += \
...
 
+ifeq ("$(BUILD_REMOTE)", "1")
+OBJS += \
+	remote_common.o \
+        remote_interface.o
+
+WORKER_OBJS := $(addprefix $(OUT)/remote/,$(filter-out remote_interface.o, $(OBJS)))
+OBJS := $(addprefix $(OUT)/, $(OBJS))
+LIBDCURL_OBJS := $(OBJS)
+OBJS += $(WORKER_OBJS)
```
marktwtn: While the `BUILD_REMOTE` is equal to 1, the variable `OBJS` is not used in the following rules.
Is the variable `LIBDCURL_OBJS` really necessary?

Makefile
```diff
@@ -139,23 +149,56 @@ OBJS += \
...
 	$(Q)$(CC) -o $@ $(CFLAGS) -I $(SRC) $(LIBTUV_INCLUDE) -c -MMD -MF $@.d $<
 
-$(OUT)/%.o: $(SRC)/%.c $(LIBTUV_PATH)/include
+$(OUT)/%.o: $(SRC)/%.c $(LIBTUV_PATH)/include $(LIBRABBITMQ_PATH)/build/include
 	$(VECHO) "  CC\t$@\n"
-	$(Q)$(CC) -o $@ $(CFLAGS) $(LIBTUV_INCLUDE) -c -MMD -MF $@.d $<
+	$(Q)$(CC) -o $@ $(CFLAGS) $(LIBTUV_INCLUDE) $(LIBRABBITMQ_INCLUDE) -c -MMD -MF $@.d $<
 
-$(OUT)/test-%: $(OUT)/test-%.o $(OBJS) $(LIBTUV_LIBRARY)
+ifeq ("$(BUILD_REMOTE)", "1")
```
marktwtn: The conditional statement can be removed if the variable `LIBDCURL_OBJS` can be replaced with `OBJS`.

ajblane: `OBJS` is for building total objects including `LIBDCURL_OBJS` and `WORKER_OBJS`. `LIBDCURL_OBJS` is for linking them for `libdcurl.so`. `WORKER_OBJS` is for linking them for remote-worker.

Makefile
ajblane: OBJS is for building total objects including LIBDCURL_OBJS and WORKER_OBJS. LIBDCURL_OBJS is for linking them for libdcurl.so. WORKER_OBJS is for linking them for remote-worker.

src/dcurl.c
```diff
@@ -96,6 +113,35 @@ int8_t *dcurl_entry(int8_t *trytes, int mwm, int threads)
     if (!isInitialized)
         return NULL;
 
+#if defined(ENABLE_REMOTE)
+
```
jserv: Remove this blank line.

jserv:
The prefix `RemoteInterface` is a bit long. We can safely rename it to `Remote` since it does not confuse.

src/remote_interface.c
```diff
@@ -243,11 +243,11 @@ RemoteInterfaceContext RemoteInterface_Context = {
     .description = "Remote interface",
     .num_max_thread = 1024,  // 1 <= num_max_thread <= AMQP_DEFAULT_MAX_CHANNELS
     .num_working_thread = 0,
-    .initialize = PoWRemoteInterface_Context_Initialize,
-    .destroy = PoWRemoteInterface_Context_Destroy,
-    .getPoWContext = PoWRemoteInterface_getPoWContext,
-    .freePoWContext = PoWRemoteInterface_freePoWContext,
+    .initialize = RI_Context_Initialize,
```
jserv: We might unify the name scheme for consistency. That is, `RI_init`, `RI_destroy`, `RI_getPoWContext`, `RI_freePoWContext`, `RI_doPoW`, etc.

Makefile
```diff
@@ -99,6 +105,10 @@ ifeq ("$(BUILD_JNI)","1")
 PREQ += $(JARS)
 endif
 
+ifeq ("$(BUILD_REMOTE)", "1")
+PREQ += $(OUT)/remote/remote-worker
```
jserv: It is not necessary to add another directory `remote`. The prefix `remote-` is self-explainable.

Makefile
```diff
@@ -139,23 +149,56 @@ OBJS += \
...
 	$(VECHO) "  LD\t$@\n"
 	$(Q)$(CC) -o $@ $^ $(LDFLAGS)
 
 $(OUT)/libdcurl.so: $(OBJS) $(LIBTUV_LIBRARY)
 	$(VECHO) "  LD\t$@\n"
 	$(Q)$(CC) -shared -o $@ $^ $(LDFLAGS)
+endif
+
+ifeq ("$(BUILD_REMOTE)", "1")
+$(OUT)/remote/%.o: $(SRC)/%.c $(LIBTUV_PATH)/include $(LIBRABBITMQ_PATH)/build/include
```
jserv: Ditto. Don't add new directory `remote`.

mk/common.mk
```diff
@@ -1,5 +1,8 @@
 # Always ensure directory build exist
 SHELL_HACK := $(shell mkdir -p $(OUT))
+ifeq ("$(BUILD_REMOTE)", "1")
+SHELL_HACK += $(shell mkdir -p $(OUT)/remote)
```
jserv: Not necessary.

mk/defs.mk
```diff
@@ -19,6 +19,9 @@ BUILD_GPU ?= 0
 # Build FPGA backend or not
 BUILD_FPGA_ACCEL ?= 0
 
+# Build remote interface backend or not
```
jserv: Its description should be `Build facilities of remote procedure calls`.

Makefile
ajblane: Okey, I try.

Makefile
```diff
@@ -139,24 +149,43 @@ OBJS += \
...
-$(OUT)/test-%: $(OUT)/test-%.o $(OBJS) $(LIBTUV_LIBRARY)
+$(OUT)/test-%: $(OUT)/test-%.o $(OBJS) $(LIBS) $(LIBTUV_LIBRARY)
 	$(VECHO) "  LD\t$@\n"
 	$(Q)$(CC) -o $@ $^ $(LDFLAGS)
 
 $(OUT)/libdcurl.so: $(OBJS) $(LIBTUV_LIBRARY)
 	$(VECHO) "  LD\t$@\n"
 	$(Q)$(CC) -shared -o $@ $^ $(LDFLAGS)
 
+ifeq ("$(BUILD_REMOTE)", "1")
```
jserv, line 177: Let's split remote interface specific build rules into a new dedicated fils in directory `mk/`.

src/remote_worker.c
```diff
@@ -0,0 +1,92 @@
...
+ * Copyright (C) 2019 dcurl Developers.
+ * Use of this source code is governed by MIT license that can be
+ * found in the LICENSE file.
+ */
+
+#include "constants.h"
+#include "dcurl.h"
+#include "remote_common.h"
+
+#define CHANNEL 1
```
jserv: Why hardcode here?

Makefile
```diff
@@ -91,6 +95,8 @@ TESTS := $(addprefix $(OUT)/test-, $(TESTS))
 LIBS = libdcurl.so
 LIBS := $(addprefix $(OUT)/, $(LIBS))
```
marktwtn: Remove the blank line.

Makefile
```diff
@@ -99,6 +105,10 @@ ifeq ("$(BUILD_JNI)","1")
 PREQ += $(JARS)
 endif
```
marktwtn: Remove the blank line.

Makefile
```diff
@@ -139,24 +149,37 @@ OBJS += \
 	pow_fpga_accel.o
 endif
 
+ifeq ("$(BUILD_REMOTE)", "1")
+OBJS += \
+	remote_common.o \
+        remote_interface.o
```
marktwtn: The indent seems weird.

mk/remote.mk
```diff
@@ -0,0 +1,8 @@
+# Build remote-worker
+$(OUT)/remote-%.o: $(SRC)/%.c $(LIBTUV_PATH)/include $(LIBRABBITMQ_PATH)/build/include
+	$(VECHO) "  CC\t$@\n"
+	$(Q)$(CC) -o $@ $(WORKER_CFLAGS) $(LIBTUV_INCLUDE) $(LIBRABBITMQ_INCLUDE) -c -MMD -MF $@.d $<
+
+$(OUT)/remote-worker: $(OUT)/remote_worker.o $(WORKER_OBJS) $(LIBS) $(LIBTUV_LIBRARY)
+	$(VECHO) "  CC\t$@\n"
```
marktwtn: Change `CC` to `LD` in the echo message.

src/dcurl.c
```diff
@@ -48,6 +51,11 @@ extern ImplContext PoWCL_Context;
 extern ImplContext PoWFPGAAccel_Context;
 #endif
 
+#if defined(ENABLE_REMOTE)
+extern RemoteInterfaceContext RemoteInterface_Context;
+static uv_sem_t notify_remoteinterface;
```
marktwtn: It is hard to read the variable name.
I think it should be simplified to `notify_remote` or at least it should be `notify_remoteInterface`.

marktwtn:
> The prefix `RemoteInterface` is a bit long. We can safely rename it to `Remote` since it does not confuse.

I saw that some functions have the name with `RemoteInterface` and the other have the prefix `RI`.
I think we should make it consistent and I prefer using `Remote` since it is more elegant and does not have to guess the meaning of the abbreviation `RI`.

src/remote_common.c
```diff
@@ -0,0 +1,337 @@
...
+
+    if (!die_on_error(amqp_basic_publish(*conn, channel, amqp_cstring_bytes(""),
+                                         amqp_cstring_bytes(queue_name), 0, 0,
+                                         &props, amqp_cstring_bytes(message)),
+                      "Publishing the message with reply_to"))
+        return false;
+
+#if defined(ENABLE_DEBUG)
+    printf(MSG_PREFIX "callback queue %s \n", (char *) props.reply_to.bytes);
+
```
jserv: Don't insert blank line here.

jserv:
Check list for merging this pull request:
1. Validate the remote interface implementation by deploying in NCKU physically.
2. Final approval by @marktwtn for design consistency.

src/remote_worker.c
```diff
@@ -0,0 +1,90 @@
...
+                   (int) envelope.message.properties.content_type.len,
+                   (char *) envelope.message.properties.content_type.bytes);
+        }
+
+#endif
+
+        /* Message body format: transacton | mwm */
+        memcpy(trytes, envelope.message.body.bytes, TRANSACTION_TRYTES_LENGTH);
+        memcpy(buf, envelope.message.body.bytes + TRANSACTION_TRYTES_LENGTH, 4);
+        sscanf(buf, "%d", &mwm);
```
jserv: The reason why sscanf might be considered bad is because it does not require you to specify maximum string width for string arguments, which could result in overflows if the input read from the source string is longer. 

Reference: https://stackoverflow.com/questions/9245682/in-c-what-is-a-safe-alternative-to-sscanf/16043903

src/remote_interface.h
```diff
@@ -0,0 +1,57 @@
...
+    /* PoW-related information */
+    PoW_Info pow_info;
+};
+
+struct _remote {
+    void *context;
+    char *description;
+    /* Connection parameters */
+    amqp_connection_state_t conn;
+    /* Multi-thread Management */
```
jserv: Simply call it "thread management" instead of "Multi-thread Management".

mk/submodule.mk
```diff
@@ -26,3 +26,19 @@ $(LIBTUV_PATH)/include:
...
+# librabbitmq related variables
+LIBRABBITMQ_PATH = deps/rabbitmq-c
+LIBRABBITMQ_INCLUDE := -I $(LIBRABBITMQ_PATH)/build/include
+LIBRABBITMQ_LIBRARY := $(LIBRABBITMQ_PATH)/build/librabbitmq/librabbitmq.a
+
+$(LIBRABBITMQ_PATH)/build/include:
+	git submodule update --init $(LIBRABBITMQ_PATH)
+	mkdir $(LIBRABBITMQ_PATH)/build
+	cd $(LIBRABBITMQ_PATH)/build && \
+         cmake -DCMAKE_INSTALL_PREFIX=. .. && \
```
jserv, line 53: @marktwtn, `cmake` is indeed one of the build dependencies. We shall mention it in documentation. Can you send pull request for it?

src/dcurl.c
```diff
@@ -96,6 +113,30 @@ int8_t *dcurl_entry(int8_t *trytes, int mwm, int threads)
...
+
+    do {
+        if (enterRemoteContext(&Remote_Context)) {
+            pow_remote_ctx = getRemoteContext(&Remote_Context, trytes, mwm);
+            goto do_remote;
+        }
+        uv_sem_wait(&notify_remote);
+    } while (1);
+
+do_remote:
```
jserv: The label `do_remote` can be renamed to `remote_pow` for consistency.

src/remote_common.c
```diff
@@ -0,0 +1,336 @@
...
+        return false;
+    }
+
+    return true;
+}
+
+bool connect_broker(amqp_connection_state_t *conn, int thread_max)
+{
+    amqp_socket_t *socket = NULL;
+    int i = 0;
```
jserv: Move the declaration of `i` to line 85 to reduce its scope.

src/remote_common.c
```diff
@@ -0,0 +1,336 @@
...
+bool declare_callback_queue(amqp_connection_state_t *conn,
+                            amqp_channel_t channel,
+                            amqp_bytes_t *reply_to_queue)
+{
+    /* Declare a exclusive private queues with TTL = 10s */
+    amqp_table_entry_t entries[1];
+    amqp_table_t table;
+    entries[0].key = amqp_cstring_bytes("x-message-ttl");
+    entries[0].value.kind = AMQP_FIELD_KIND_U32;
+    entries[0].value.value.u32 = 10000;  // 10s
```
jserv: How about changing its value to `10 * 1000`?

mk/submodule.mk
marktwtn: Sure~

src/dcurl.c
```diff
@@ -96,6 +113,30 @@ int8_t *dcurl_entry(int8_t *trytes, int mwm, int threads)
...
+    do {
+        if (enterRemoteContext(&Remote_Context)) {
+            pow_remote_ctx = getRemoteContext(&Remote_Context, trytes, mwm);
+            goto do_remote;
+        }
+        uv_sem_wait(&notify_remote);
+    } while (1);
+
+remote_pow:
+    if (!doRemoteContext(&Remote_Context, pow_remote_ctx)) {
```
jserv: It is ambiguous to figure out two variables with similar names: `Remote_Context` and `pow_remote_ctx`.
Can you simply say `remote_ctx` and `pow_ctx`? The later is just a local variable.

ajblane: I maybe rename `Remote_Context` to `Remote_ImplConext`. In the future, we maybe use ZeroMQ to Implement RPC or use others and will replace `pw_remote_ctx` with `pow_ctx`.

jserv: It sounds good.

jserv:
Update `docs/` and entries in `README.md` accordingly.

src/remote_worker.c
```diff
@@ -0,0 +1,90 @@
...
+        memset(buf, '0', sizeof(buf));
+
+#if defined(ENABLE_DEBUG)
+        printf(MSG_PREFIX "PoW is done\n");
+#endif
+        if (!acknowledge_broker(&conn, 1, &envelope))
+            goto fail_to_exit;
+
+#if defined(ENABLE_DEBUG)
+        printf(MSG_PREFIX "Sending a ack is done\n");
```
jserv: Fix the grammar. Should be "an" ACK.

jserv:
Travis-CI indicated the absence of OpenSSL:
> Could NOT find OpenSSL

Please attempt to fix it before merging. Ask @ender503 for macOS environment trial.

jserv (changes requested):
Travis-CI still complained. Please validate.

ajblane:
> Travis-CI indicated the absence of OpenSSL:
> 
> > Could NOT find OpenSSL
> 
> Please attempt to fix it before merging. Ask @ender503 for macOS environment trial.

Thank @ender503 for this help. Solve it to refer to https://github.com/tarantool/tarantool/issues/1852.

src/remote_common.c
```diff
@@ -0,0 +1,336 @@
...
+    /*Wait for three frames: AMQP_FRAME_METHOD, AMQP_FRAME_HEADER,
+     * AMQP_FRAME_BODY*/
+    for (;;) {
+        amqp_maybe_release_buffers(*conn);
+        amqp_simple_wait_frame(*conn, &frame);
+        if (!die_on_amqp_error(amqp_get_rpc_reply(*conn), "Wait method frame"))
+            return false;
+
+#if defined(ENABLE_DEBUG)
+        printf(MSG_PREFIX "Frame type: %u channel: %u\n", frame.frame_type, frame.channel);
```
jserv: We might introduce new inline function/macro `dprintf` which follows the presence of `ENABLE_DEBUG` to wrap `printf`. @marktwtn, any idea?

mk/submodule.mk
marktwtn: I will send a pull request after this one is merged.

src/remote_worker.c
```diff
@@ -0,0 +1,90 @@
...
+            (unsigned) envelope.delivery_tag, (int) envelope.exchange.len,
+            (char *) envelope.exchange.bytes, (int) envelope.routing_key.len,
+            (char *) envelope.routing_key.bytes,
+            (char *) envelope.message.properties.reply_to.bytes);
+        if (envelope.message.properties._flags & AMQP_BASIC_CONTENT_TYPE_FLAG) {
+            printf(MSG_PREFIX "Content-type: %.*s\n",
+                   (int) envelope.message.properties.content_type.len,
+                   (char *) envelope.message.properties.content_type.bytes);
+        }
+
```
marktwtn, line 44: remove the blank line.

src/remote_interface.c
```diff
@@ -0,0 +1,239 @@
...
+    return true;
+
+fail_to_validation:
+    freeTrobject(ret_trits);
+    freeTrobject(hash_trytes);
+fail_to_hashtrytes:
+    freeTrobject(trytes_t);
+fail_to_inittrytes:
+    return false;
+}
```
marktwtn, line 96: Please append a blank line.

Makefile
```diff
@@ -139,24 +147,37 @@ OBJS += \
...
 	$(Q)$(CC) -o $@ $(CFLAGS) -I $(SRC) $(LIBTUV_INCLUDE) -c -MMD -MF $@.d $<
 
-$(OUT)/%.o: $(SRC)/%.c $(LIBTUV_PATH)/include
+$(OUT)/%.o: $(SRC)/%.c $(LIBTUV_PATH)/include $(LIBRABBITMQ_PATH)/build/include
 	$(VECHO) "  CC\t$@\n"
-	$(Q)$(CC) -o $@ $(CFLAGS) $(LIBTUV_INCLUDE) -c -MMD -MF $@.d $<
+	$(Q)$(CC) -o $@ $(CFLAGS) $(LIBTUV_INCLUDE) $(LIBRABBITMQ_INCLUDE) -c -MMD -MF $@.d $<
 
-$(OUT)/test-%: $(OUT)/test-%.o $(OBJS) $(LIBTUV_LIBRARY)
+$(OUT)/test-%: $(OUT)/test-%.o $(OBJS) $(LIBS) $(LIBTUV_LIBRARY)
```
marktwtn: I was thinking about the consistency.
The `rabbitmq` library is added into the `$(LIBS)` but `libtuv` library isn't.
I think we should modify one of them to make is consistent.

src/remote_interface.h
```diff
@@ -0,0 +1,57 @@
...
+
+#include <stdbool.h>
+#include <stdint.h>
+#include "common.h"
+#include "constants.h"
+#include "remote_common.h"
+#include "uv.h"
+
+typedef struct _pow_remote_context PoW_Remote_Context;
+typedef struct _remote_impl_context RemoteImplContext;
```
marktwtn, line 14: The naming convention seems different.

src/remote_common.c
marktwtn: How about adding these macros in `src/common.h`:
```
#if defined(ENABLE_DEBUG)
    #define DPRINTF(...) printf(__VA_ARGS__)
#else
    #define DPRINTF(...)
#endif
```

Makefile
ajblane: I will use `$(LIBRABBITMQ_LIBRARY)` rather than `$(LIBS)`.

src/remote_interface.h
ajblane: Yes, they are dfferent, as refers to https://github.com/DLTcollab/dcurl/blob/develop/src/pow_avx.h#L11, https://github.com/DLTcollab/dcurl/blob/develop/src/implcontext.h#L10.

src/dcurl.c
```diff
@@ -48,6 +51,11 @@ extern ImplContext PoWCL_Context;
 extern ImplContext PoWFPGAAccel_Context;
 #endif
 
+#if defined(ENABLE_REMOTE)
+extern RemoteImplContext Remote_ImplContext;
```
marktwtn: I think we should change the name from `Remote_ImplContext` to `Remote_Context`.
Just like the naming of the previous delcared `ImplContext` structure variable.

src/remote_interface.h
marktwtn: Got it.
I guess we will need a consistent structure naming convention in the future.

mk/submodule.mk
marktwtn: I'll start to prepare the pull request since `libtuv` requires `cmake`, too.

src/dcurl.c
marktwtn: Change the variable name `Remote_ImplContext` to `Remote_Context` for consistency.

mk/submodule.mk
marktwtn: Done in the commit 07fa452.

Makefile
ajblane: I have built this PR on de10-nano and there is a linking error,  `could not read symbols: Bad value`. It seems to need to recompile RabbitMQ with `-fPIC`, as refers to https://github.com/rust-lang/cargo/issues/713.

The current workaround is by defining `$(LIBTUV_LIBRARY_LINK)`
```
$(OUT)/test-%: $(OUT)/test-%.o $(OBJS) $(LIBTUV_LIBRARY)
     $(VECHO) "  LD\t$@\n"
     $(Q)$(CC) -o $@ $^ $(LDFLAGS) $(LIBTUV_LIBRARY_LINK)
```
```
$(LIBTUV_LIBRARY_LINK) = -Wl,-Bdynamic -L$(LIBRABBITMQ_PATH)/build/librabbitmq/ -llibrabbitmq
```
Any suggestion?

jserv:
You are about to merge this pull request. Please rebase on top of latest `develop` branch

src/dcurl.c
ajblane: How about renaming `RabbitMQ_Context`? This is because `Remote_ImplContext` can be implemented by using ZeroMq or others.

marktwtn: Do we need to execute multiple message queue implementation at the same time?
If not, I guess using `Remote_Context` is fine.
Since compare to the file `src/pow_c.c`, the variable `PoWC_Context` indicates the PoW implementation rather than the message queue implementation.
It might be a little confusing if we use `RabbitMQ_Context`.

Makefile
marktwtn: The `libtuv` is compiled with the flag `-fPIC`, too.

ajblane: Do you mean we need to fork RabbitMQ C client into DLTcollab in order to add `-fPIC` flag and maintain   it?

src/dcurl.c
ajblane: How about renaming `RemoteRabbitmq_Context`?

ajblane:
Fix the CI error now

ajblane:
> Fix the CI error now

Thank @ender503  for testing on macOS and discussing.

jserv:
Since #154 is landed, `ddprintf` should be used diversely.

src/remote_common.c
```diff
@@ -0,0 +1,336 @@
...
+#include <string.h>
+
+bool die_on_amqp_error(amqp_rpc_reply_t x, char const *context)
+{
+    switch (x.reply_type) {
+    case AMQP_RESPONSE_NORMAL:
+        return true;
+
+    case AMQP_RESPONSE_NONE:
+        fprintf(stderr, "%s: missing RPC reply type!\n", context);
```
jserv: We shall not use `stderr` because `dcurl` is intended to be a library bridged to existing applications. Instead, `ddprintf` is a wrapper for logging and diagnosis purpose.

marktwtn:
There are too many commits in this pull request.
If the commit modification is something like blank line, typo, and variable renaming which is caused by the first commit, they should be merged into the first commit.

jserv:
> There are too many commits in this pull request.
> If the commit modification is something like blank line, typo, and variable renaming which is caused by the first commit, they should be merged into the first commit.

As @marktwtn mentioned, @ajblane should use `git rebase -i` to rework these commits.

Makefile
```diff
@@ -91,6 +95,7 @@ TESTS := $(addprefix $(OUT)/test-, $(TESTS))
 
 LIBS = libdcurl.so
 LIBS := $(addprefix $(OUT)/, $(LIBS))
+LIBS += $(LIBRABBITMQ_LIBRARY)
```
marktwtn: What is the value of this variable?
I do not see it from the source code.

Makefile
```diff
@@ -140,23 +147,36 @@ OBJS += \
 	pow_fpga_accel.o
 endif
 
+ifeq ("$(BUILD_REMOTE)", "1")
+OBJS += \
+	remote_common.o \
+	remote_interface.o
+
+WORKER_OBJS := $(addprefix $(OUT)/remote-,$(filter-out remote_interface.o, $(OBJS)))
```
marktwtn: The `$(OUT)/remote` can be changed to `$(OUT)/worker` to avoid naming confusion like **remote-remote_common.o**.

Makefile
```diff
@@ -140,23 +147,36 @@ OBJS += \
...
 	$(VECHO) "  CC\t$@\n"
 	$(Q)$(CC) -o $@ $(CFLAGS) -I $(SRC) $(LIBTUV_INCLUDE) -c -MMD -MF $@.d $<
 
-$(OUT)/%.o: $(SRC)/%.c $(LIBTUV_PATH)/include
+$(OUT)/%.o: $(SRC)/%.c $(LIBTUV_PATH)/include $(LIBRABBITMQ_PATH)/build/include
 	$(VECHO) "  CC\t$@\n"
-	$(Q)$(CC) -o $@ $(CFLAGS) $(LIBTUV_INCLUDE) -c -MMD -MF $@.d $<
+	$(Q)$(CC) -o $@ $(CFLAGS) $(LIBTUV_INCLUDE) $(LIBRABBITMQ_INCLUDE) -c -MMD -MF $@.d $<
 
 $(OUT)/test-%: $(OUT)/test-%.o $(OBJS) $(LIBTUV_LIBRARY)
```
marktwtn: Please add the dependency of rabbitMQ library.

Makefile
```diff
@@ -140,23 +147,36 @@ OBJS += \
...
 	$(VECHO) "  LD\t$@\n"
-	$(Q)$(CC) -o $@ $^ $(LDFLAGS)
+	$(Q)$(CC) -o $@ $^ $(LDFLAGS) $(LIBRABBITMQ_LINK)
 
 $(OUT)/libdcurl.so: $(OBJS) $(LIBTUV_LIBRARY)
```
marktwtn: Ditto.

mk/remote.mk
```diff
@@ -0,0 +1,8 @@
+# Build remote-worker
+$(OUT)/remote-%.o: $(SRC)/%.c $(LIBTUV_PATH)/include $(LIBRABBITMQ_PATH)/build/include
```
marktwtn: This should be modified to `worker`, too.

mk/remote.mk
```diff
@@ -0,0 +1,8 @@
...
+	$(VECHO) "  CC\t$@\n"
+	$(Q)$(CC) -o $@ $(WORKER_CFLAGS) $(LIBTUV_INCLUDE) $(LIBRABBITMQ_INCLUDE) -c -MMD -MF $@.d $<
+
+$(OUT)/remote-worker: $(OUT)/remote_worker.o $(WORKER_OBJS) $(LIBTUV_LIBRARY)
```
marktwtn: Please add the dependency of rabbitMQ library.

mk/submodule.mk
```diff
@@ -26,3 +26,34 @@ $(LIBTUV_PATH)/include:
...
+    LIBRABBITMQ_LINK := -Wl,-rpath=$(LIBRABBITMQ_LIB_PATH) -L$(LIBRABBITMQ_LIB_PATH) -lrabbitmq
+    LIBRABBITMQ_LIBRARY := $(LIBRABBITMQ_LIB_PATH)/librabbitmq.so
+endif
+
+$(LIBRABBITMQ_PATH)/build/include:
+	git submodule update --init $(LIBRABBITMQ_PATH)
+	mkdir $(LIBRABBITMQ_PATH)/build
+ifeq ($(UNAME_S),darwin)
+	# macOS
+	cd $(LIBRABBITMQ_PATH)/build && \
```
marktwtn, line 48: I wonder whether this is really necessary to enter the `$(LIBRABBITMQ_PATH)/build` directory.
It seems using `$(LIBRABBITMQ_PATH)` is fine.

docs/remote-interface.md
```diff
@@ -0,0 +1,67 @@
...
+ |            |                     ^            |
+ +------------|---------------------|------------+
+              v                     |
+   +---------------------------------------------+
+  +---------------------------------------------+|
+ +---------------------------------------------+|+
+ |                  remote worker              |+
+ +---------------------------------------------+
+```
+To support asynchronouse remote procedure call, remote interface, `Remote_ImplContext`, in dcurl provides an interface to implement it. dcurl currently uses RabbitMQ C client to implement asynchronouse RPC in remote interface. Remote interface provides thread management to can run an asynchronouse RPC per thread.
```
marktwtn: Typo `asynchronous`.
And it should be
```
To support asynchronous remote procedure call and remote interface, `Remote_ImplContext` in dcurl ......
```

`Remote interface provides thread management to can run an asynchronouse`?
Another typo and the `to can run` seems weird.

src/remote_worker.c
```diff
@@ -0,0 +1,77 @@
...
+                (char *) ret_trytes))
+            goto fail_to_exit;
+
+        free(ret_trytes);
+        amqp_destroy_envelope(&envelope);
+        ddprintf(MSG_PREFIX "Publishing PoW result to callback queue is done\n");
+        ddprintf(MSG_PREFIX "---\n");
+    }
+
+fail_to_exit:
```
marktwtn: We should name it as `fail` only since `fail to` has the meaning of unable to do something.

src/remote_worker.c
```diff
@@ -0,0 +1,77 @@
...
+    dcurl_destroy();
+
+    return 0;
```
marktwtn: It is weird that if the program failed, it returned 0.
You might reference the correct way of returning value from the source code of [pow_c.c](https://github.com/DLTcollab/dcurl/blob/develop/src/pow_c.c).

src/remote_interface.h
```diff
@@ -0,0 +1,59 @@
...
+
+struct _remote_impl_context {
+    void *context;
+    char *description;
+    /* Connection parameters */
+    amqp_connection_state_t conn[CONN_MAX];
+    /* Thread management */
+    uv_mutex_t lock;
+    bool slots[CONN_MAX]; /* Used to tell which slot is
+                                              available */
```
marktwtn, line 35: The typesetting of the comment is weird.
Did you use the **clang-format** to check the style?

ajblane: yes

src/remote_worker.c
marktwtn: Or just like what you did in the other part of this pull request.

src/remote_common.c
```diff
@@ -0,0 +1,325 @@
...
+                                "Opennng the channel")))
+        goto fail_to_open_channel;
+
+    return true;
+
+fail_to_open_channel:
+    for (int j = 0; j < i; j++)
+        amqp_channel_close(*conn, j + 1, AMQP_REPLY_SUCCESS);
+
+fail_to_login_error:
```
marktwtn: `fail_to_login` sounds more reasonable.

src/remote_common.c
```diff
@@ -0,0 +1,325 @@
...
+    amqp_connection_close(*conn, AMQP_REPLY_SUCCESS);
+
+fail_to_broker_close:
```
marktwtn: Should be renamed.

src/remote_common.c
```diff
@@ -0,0 +1,325 @@
...
+
+        if (frame.frame_type != AMQP_FRAME_METHOD)
+            continue;
+
+        ddprintf(MSG_PREFIX "Method: %s\n", amqp_method_name(frame.payload.method.id));
+
+        if (frame.payload.method.id != AMQP_BASIC_DELIVER_METHOD)
+            continue;
+
+#if defined(ENABLE_DEBUG)
```
marktwtn, line 203: The `ENABLE_DEBUG` can be removed since you have already used the `ddprintf()`.
At least `ddprintf()` should not be placed between the `#if defined(ENABLE_DEBUG) ... #endif` macro.

src/remote_common.c
```diff
@@ -0,0 +1,325 @@
...
+
+        if (frame.frame_type != AMQP_FRAME_HEADER) {
+            ddprintf("Unexpected header!");
+            return false;
+        }
+
+#if defined(ENABLE_DEBUG)
+        p = (amqp_basic_properties_t *) frame.payload.properties.decoded;
+        if (p->_flags & AMQP_BASIC_CONTENT_TYPE_FLAG) {
+            ddprintf(MSG_PREFIX "Content-type: %.*s\n", (int) p->content_type.len,
```
marktwtn, line 224: Ditto.

mk/submodule.mk
ajblane: I think that using `cmake`  keeps simple as entering $(LIBRABBITMQ_PATH)/build directory to build and configure RabbitMQ.

marktwtn: For example, can't it be like:
```
cmake -DOPENSSL_ROOT_DIR=/usr/local/opt/openssl/ -DCMAKE_INSTALL_PREFIX=./build . && \
cmake --build ./build --target install
```

ajblane: An error occurs as entering $(LIBRABBITMQ_PATH) and using the above command to build as follows:
```
-- Found POPT: /usr/include
-- Found XMLTO: /usr/bin/xmlto
-- Could NOT find Doxygen (missing: DOXYGEN_EXECUTABLE)
-- Building rabbitmq as a shared library - yes
-- Building rabbitmq as a static library - yes
-- Configuring done
-- Generating done
-- Build files have been written to: /home/xxxx/Desktop/dcurl/deps/rabbitmq-c
Error: could not load cache
```

src/remote_common.c
ajblane: `p` is defined and used by `ddprintf` when `make BUILD_DEBUG`. Therefore, the `ENABLE_DEBUG` needs to be used.

PR #147, opened 2019-05-06
docs: Revise description of IRI adaptation

marktwtn:


README.md
```diff
@@ -30,16 +30,17 @@ After integrating dcurl into IRI, performance of [attachToTangle](https://iota.r
...
+* ```$ cd ~/iri && make check```
 * ```$ java -jar target/iri.jar -p <port> --pearldiver-exlib dcurl```
 
-or with the **deprecated** commands
+or use the **deprecated** commands to load from the copied library file
 
-* ```$ cd ~/iri && mvn compile && mvn package```
+* ```$ cd ~/iri && make check```
 * ```$ cp ~/dcurl/build/libdcurl.so ~/iri```
 * ```$ cd ~/iri && java -Djava.library.path=./ -jar target/iri.jar -p <port> --pearldiver-exlib dcurl```
```
jserv: Is this line necessary now? We don't have to rely on the setting of `java.library.path`.

marktwtn: It is the old way to use IRI + dcurl.
The IRI still supports it and I marked it as deprecated commands.

jserv: It is confusing to me.

marktwtn: We can remove the deprecated commands and description in the document.
I just keep them temporarily and will remove them in the future, for backward JNI support?

https://github.com/DLTcollab/dcurl/pull/141#discussion_r277945396

jserv: I agree with the removal of deprecated commands. However, you can even insert new section for trouble shooting where backward JNI would be mentioned.

marktwtn: I removed the deprecated commands.
However, I did not add anything about the trouble shooting in the document.

Reason:
The generated `iri.jar` can not be completed if the `dcurl` JAR file is not installed.
If the `dcurl` JAR file is installed correctly, then everything will be fine.
If it doesn't, then the final `iri.jar` can not be generated. The deprecated commands would not help in this kind of situation.

PR #154, opened 2019-05-13
refactor: Simplification of output debug messages

### LINKED ISSUE #153: Introduce macro for printing the debug messages
Base on the discussion of https://github.com/DLTcollab/dcurl/pull/143#discussion_r281890019, the way to print the debug messages should be simplified.
The original code
```
#if defined(ENABLE_DEBUG)
    printf(......);
#endif
```
would be replaced with the macro
```
DPRINTF(...);
```

The `DPRINTF(...)` macro implementation is
```
#if defined(ENABLE_DEBUG)
    #define DPRINTF(...) printf(__VA_ARGS__)
#else
    #define DPRINTF(...)
#endif
```
It will be written in the `src/common.h` and allow other `dcurl` source code to use it.

marktwtn:
Close #153.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -1,7 +1,9 @@
 #ifndef COMMON_H_
 #define COMMON_H_
 
+#include <stdarg.h>
 #include <stdint.h>
+#include <stdio.h>
 #include <time.h>
 
 #define __DCURL_MAJOR__ 0
@@ -9,6 +11,15 @@
 #define __DCURL_PATCH__ 0
 
 double diff_in_second(struct timespec t1, struct timespec t2);
+
+static inline void ddprintf(const char *format, ...) {
+#if defined(ENABLE_DEBUG)
+    va_list ap;
+    va_start(ap, format);
+    vprintf(format, ap);
+    va_end(ap);
+#endif
+}
 
 typedef struct _pow_info PoW_Info;
 
--- a/src/implcontext.c
+++ b/src/implcontext.c
@@ -14,12 +14,10 @@ bool registerImplContext(ImplContext *impl_ctx)
 bool initializeImplContext(ImplContext *impl_ctx)
 {
     bool res = impl_ctx->initialize(impl_ctx);
-#if defined(ENABLE_DEBUG)
     if (res) {
-        printf(MSG_PREFIX "Implementation %s is initialized successfully\n",
-               impl_ctx->description);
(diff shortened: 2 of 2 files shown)


src/common.h
```diff
@@ -8,6 +8,12 @@
 #define __DCURL_MINOR__ 1
 #define __DCURL_PATCH__ 0
 
+#if defined(ENABLE_DEBUG)
+#define DPRINTF(...) printf(__VA_ARGS__)
```
jserv: Eventually, `dprintf` might be conjuncted with logging facilities such as something in `entangle`. I would expect there is no direct call to `printf`.

jserv: Lower-case `dprintf` is better than the opposite form `DPRINTF`.

wusyong (approved):

src/common.h
marktwtn: > Eventually, `dprintf` might be conjuncted with logging facilities such as something in `entangle`. I would expect there is no direct call to `printf`.

I guess it will be left to the future modification?

jserv: We can proceed at the moment since it could be involved in small chunks of changes.

marktwtn: I found the logging function of `entangled`: [log_debug()](https://github.com/iotaledger/entangled/blob/09f58f17072053798b0d4adec0b1263eb005641c/utils/logger_helper.h#L23)

```c
static inline void dprintf(...) {
#if defined(ENABLE_DEBUG)
    log_debug();
#endif
}

void log_debug() {
    // The current dcurl implementation would be printf()
    // In the future,
    // it might be integrated with other logging facilities like entangled
    printf(...);
}
```

Is this what you expected to see?

jserv: `selftest` programs inside Linux kernel tree come with a nice example for writing dprintf: https://github.com/torvalds/linux/blob/master/tools/testing/selftests/x86/pkey-helpers.h

marktwtn: Since `dprintf()` has been declared in **stdio.h**, I rename it to `dbgprintf()` instead.

src/common.h
```diff
@@ -10,6 +12,15 @@
 
 double diff_in_second(struct timespec t1, struct timespec t2);
 
+static inline void dbgprintf(const char *format, ...) {
```
jserv: How about renaming to `ddprintf`, which means dcurl debug-oriented printf?

marktwtn: Agree.

PR #156, opened 2019-05-20
perf: Optimize trits from trytes with x86 SIMD

### LINKED ISSUE #92: Optimize Trits/Trytes manipulation with SIMD techniques

marktwtn:
Without SIMD optimization
Input size(byte) - Average time(nsec)
81               - 355.6
2592             - 5752.3
2673             - 6273.0

With SIMD optimization
Input size(byte) - Average time(nsec)
81               - 167.1
2592             - 1751.8
2673             - 2098.8

Hardware information
architecture - x86_64
CPU          - AMD Ryzen 5 2400G

Close #92.

### DIFF
--- a/src/trinary_sse42.h
+++ b/src/trinary_sse42.h
@@ -247,4 +247,201 @@ static inline Trobject_t *trytes_from_trits_sse42(Trobject_t *trits)
     return trytes;
 }
 
+static inline Trobject_t *trits_from_trytes_sse42(Trobject_t *trytes)
+{
+    Trobject_t *trits = NULL;
+    int8_t *src = (int8_t *) malloc(trytes->len * 3);
+
+    const int block_8bit = BLOCK_8BIT(__m128i);
+    /* For setting the most significant bit of a byte */
+    const int8_t setMSB = 0x80;
+    static int8_t TrytesToTritsMappings[][3] = {
+        {0, 0, 0},   {1, 0, 0},   {-1, 1, 0},  {0, 1, 0},   {1, 1, 0},
+        {-1, -1, 1}, {0, -1, 1},  {1, -1, 1},  {-1, 0, 1},  {0, 0, 1},
+        {1, 0, 1},   {-1, 1, 1},  {0, 1, 1},   {1, 1, 1},   {-1, -1, -1},
+        {0, -1, -1}, {1, -1, -1}, {-1, 0, -1}, {0, 0, -1},  {1, 0, -1},
+        {-1, 1, -1}, {0, 1, -1},  {1, 1, -1},  {-1, -1, 0}, {0, -1, 0},
+        {1, -1, 0},  {-1, 0, 0}};
+    /* The set and range for indicating the trits value (0, 1, -1)
+     * of the corresponding trytes */
+    /* '9', 'C', 'F', 'I', 'L', 'O', 'R', 'U', 'X' */
+    const char *setLowTrit0 = "9CFILORUX";
+    /* 'A', 'D', 'G', 'J', 'M', 'P', 'S', 'V', 'Y' */
+    const char *setLowTritP1 = "ADGJMPSVY";
+    /* 'B', 'E', 'H', 'K', 'N', 'Q', 'T', 'W', 'Z' */
+    const char *setLowTritN1 = "BEHKNQTWZ";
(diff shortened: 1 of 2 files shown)


src/trinary_sse42.h
```diff
@@ -247,4 +247,201 @@ static inline Trobject_t *trytes_from_trits_sse42(Trobject_t *trits)
     return trytes;
 }
 
+static inline Trobject_t *trits_from_trytes_sse42(Trobject_t *trytes)
+{
+    Trobject_t *trits = NULL;
+    int8_t *src = (int8_t *) malloc(trytes->len * 3);
```
jserv, line 253: It might benefit from memory pool. Let's rework later.

marktwtn: Got it.

PR #158, opened 2019-05-22
test: Add thread-safe test

### LINKED ISSUE #151: Allow dcurl to test thread-safe
In the past, for dcurl, it supports thread-safe test with python, as refers to [test-multi-pow.py]( https://github.com/DLTcollab/dcurl/blob/a379862ab0710401735323c7515b18b45d96daee/tests/test-multi-pow.py). This issue can be expected to allow dcurl to test thread-safe using C.

ajblane:
Resolved #151


tests/test-multi-pow.c
```diff
@@ -0,0 +1,107 @@
+/* Test program for thread-safe dcurl */
+#include <pthread.h>
+#include "common.h"
+#include "dcurl.h"
+
+#define THREAD_MAX 10
+
+typedef struct _ducrl_item DCURL_ITEM;
```
jserv: Use lower case form, such as `dcurl_item`. The uppercase might imply macros or constants.

tests/test-multi-pow.c
```diff
@@ -0,0 +1,107 @@
...
+struct _ducrl_item {
+    int mwm;
+    int8_t input_trytes[TRANSACTION_TRYTES_LENGTH];  /* 2673 */
+    int8_t output_trytes[TRANSACTION_TRYTES_LENGTH]; /* 2673 */
+};
+
+void *dcurl_entry_cb(void *arg)
+{
+    DCURL_ITEM *dcurl_item = (DCURL_ITEM *) arg;
```
jserv: Ditto. Variable name can be simply `item`.

tests/test-multi-pow.c
```diff
@@ -0,0 +1,107 @@
...
+    /* test dcurl Implementation with mwm = 9 */
+    int8_t *ret_trytes =
+        dcurl_entry(dcurl_item->input_trytes, dcurl_item->mwm, 0);
+    assert(ret_trytes);
```
jserv: You can append messages with assert macro. e.g. `assert(a == b && "A is not equal to B");`

tests/test-multi-pow.c
```diff
@@ -0,0 +1,107 @@
...
+    memcpy(dcurl_item->output_trytes, ret_trytes, TRANSACTION_TRYTES_LENGTH);
+    free(ret_trytes);
+
+    pthread_exit(NULL);
```
jserv, line 25: I am not sure if it is safe to invoke `pthread_exit` here.

jserv:
I am afraid of the limited coverage for concurrency tests. It might not reflect that we suffer from in real-world usages.

tests/test-multi-pow.c
```diff
@@ -0,0 +1,107 @@
...
+
+    for (int i = 0; i < THREAD_MAX; i++) {
+        Trytes_t *trytes_t =
+            initTrytes(dcurl_items[i].output_trytes, TRANSACTION_TRYTES_LENGTH);
+        assert(trytes_t);
+        Trytes_t *hash_trytes = hashTrytes(trytes_t);
+        assert(hash_trytes);
+
+        /* Validation */
+        Trits_t *ret_trits = trits_from_trytes(hash_trytes);
```
marktwtn, line 95: Please remember to release the source of `ret_trits`.

tests/test-multi-pow.c
marktwtn: Hmm...... isn't it the right way of terminating a thread?
Is there any potential risk?

jserv: We might have no information how the program is terminated, that is a matter of test plan flow control.

jserv: Eventually, we would not depend on the messages. Instead, we shall be able to check all return values to be acquainted with the state of given test items.

tests/test-multi-pow.c
```diff
@@ -0,0 +1,108 @@
...
+        "99999999999999999A9RGRKVGWMWMKOLVMDFWJUHNUNYWZTJADGGPZGXNLERLXYWJE9WQH"
+        "WWBMCPZMVVMJUMWWBLZLNMLDCGDJ999999999999999999999999999999999999999999"
+        "999999999999YGYQIVD99999999999999999999TXEFLKNPJRBYZPORHZU9CEMFIFVVQBU"
+        "STDGSJCZMBTZCDTTJVUFPTCCVHHORPMGCURKTH9VGJIXUQJVHK99999999999999999999"
+        "9999999999999999999999999999999999999999999999999999999999999999999999"
+        "9999999999999999999999999999999999999999999999999999999999999999999999"
+        "9999999999999999999999999999999999999999999999999999999999999999999999"
+        "9999999999999";
+
+    int mwm = 9;
```
marktwtn: This should be changed to 14.
9 is for the testnet and 14 is for the mainnet.

You might copy it from the other test case.
However, I do not know why the `test-dcurl.c` uses 9 instead of 14.

tests/test-multi-pow.c
```diff
@@ -0,0 +1,108 @@
...
+                       (void *) &items[i]);
+    }
+
+    for (int i = 0; i < THREAD_MAX; i++)
+        pthread_join(threads[i], NULL);
+
+    for (int i = 0; i < THREAD_MAX; i++) {
+        Trytes_t *trytes_t =
+            initTrytes(items[i].output_trytes, TRANSACTION_TRYTES_LENGTH);
+        assert(trytes_t && "initTrytes() fails");
```
marktwtn: I think we should change the `fails` to `failed`.

tests/test-multi-pow.c
```diff
@@ -0,0 +1,108 @@
...
+        Trytes_t *hash_trytes = hashTrytes(trytes_t);
+        assert(hash_trytes && "hashTrytes() fails");
```
marktwtn: Ditto.

tests/test-multi-pow.c
```diff
@@ -0,0 +1,108 @@
...
+
+        /* Validation */
+        Trits_t *ret_trits = trits_from_trytes(hash_trytes);
+        for (int j = 243 - 1; j >= 243 - items[i].mwm; j--) {
+            assert(ret_trits->data[j] == 0 && "Validation fails");
```
marktwtn: Ditto.

tests/test-multi-pow.c
```diff
@@ -0,0 +1,108 @@
...
+    int8_t output_trytes[TRANSACTION_TRYTES_LENGTH]; /* 2673 */
+};
+
+void *dcurl_entry_cb(void *arg)
+{
+    dcurl_item *item = (dcurl_item *) arg;
+    /* test dcurl Implementation with mwm = 9 */
+    int8_t *ret_trytes =
+        dcurl_entry(item->input_trytes, item->mwm, 0);
+    assert(ret_trytes && "dcurl_entry() fails");
```
marktwtn: Change the `fails` to `failed`.

tests/test-multi-pow.c
```diff
@@ -0,0 +1,108 @@
...
+struct _dcurl_item {
+    int mwm;
+    int8_t input_trytes[TRANSACTION_TRYTES_LENGTH];  /* 2673 */
+    int8_t output_trytes[TRANSACTION_TRYTES_LENGTH]; /* 2673 */
+};
+
+void *dcurl_entry_cb(void *arg)
+{
+    dcurl_item *item = (dcurl_item *) arg;
+    /* test dcurl Implementation with mwm = 9 */
```
marktwtn: The comment should be modified to 14.

PR #162, opened 2019-06-05
refactor: Change blocked RPC to non-blocked RPC

### LINKED ISSUE #161: Remote interface: Change into non-blocked RPC
In the case of all remote workers which are not workable, blocked PRCs not get RPC results until remote workers are workable. Therefore, remote interface needs to change blocked RPC to non-blocked RPC with 10s. When time-out occurs, remote interface goes to run local dcurl.

ajblane:
In the case of all remote workers which are not workable, blocked PRCs
not get RPC results until remote workers are workable. Therefore,
remote interface needs to change blocked RPC to non-blocked RPC with
10s. When time-out occurs, remote interface goes to run local dcurl.

Pass test with a broker and remote interface without remote worker.

Close #161

### DIFF
--- a/src/remote_common.c
+++ b/src/remote_common.c
@@ -7,6 +7,7 @@
 #include "remote_common.h"
 #include <stdio.h>
 #include <string.h>
+#include <sys/time.h>
 #include "common.h"
 
 bool die_on_amqp_error(amqp_rpc_reply_t x, char const *context)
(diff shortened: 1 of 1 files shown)


src/remote_common.c
```diff
@@ -186,7 +187,11 @@ bool wait_response_message(amqp_connection_state_t *conn,
      * AMQP_FRAME_BODY*/
     for (;;) {
         amqp_maybe_release_buffers(*conn);
-        amqp_simple_wait_frame(*conn, &frame);
+
+        struct timeval t = {10, 0}; /* RPC timeout: 10s*/
```
wusyong, line 191: Just curious why 10 seconds is chosen. And maybe these could be macros?

jserv:
Don't change the lines related to `ddprintf`. Let's improve diagnosis message in another pull request. In this RP, we shall refactor.

src/remote_common.c
```diff
@@ -186,7 +187,11 @@ bool wait_response_message(amqp_connection_state_t *conn,
...
+        if(!die_on_error(amqp_simple_wait_frame_noblock(*conn, &frame, &t), "RPC timeout"))
```
jserv: Did you use `clang-format -i` to indent?

src/remote_common.c
ajblane: It can be changed to other values to make sense and I don't know much about that. 
For consistent, MACRO does not be used.

wusyong (approved):

PR #164, opened 2019-06-06
test: Modify the difficulty and resource of testcases

marktwtn:
Change the mwm from 9(testnet) to 14(mainnet).
Use the fixed number of threads rather than the maximum available threads,
since the testcases are for validating the dcurl functionality,
not for exhausting the hardware resource.

### DIFF
--- a/tests/test-dcurl.c
+++ b/tests/test-dcurl.c
@@ -47,12 +47,12 @@ int main()
         "9999999999999999999999999999999999999999999999999999999999999999999999"
         "9999999999999";
 
-    int mwm = 9;
+    int mwm = 14;
 
     for (int loop_count = 0; loop_count < LOOP_MAX; loop_count++) {
-        /* test dcurl Implementation with mwm = 9 */
+        /* test dcurl Implementation with mwm = 14 */
         dcurl_init();
-        int8_t *ret_trytes = dcurl_entry((int8_t *) trytes, mwm, 0);
+        int8_t *ret_trytes = dcurl_entry((int8_t *) trytes, mwm, 8);
         assert(ret_trytes);
         dcurl_destroy();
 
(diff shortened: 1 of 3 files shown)


marktwtn:
I do not know why the commit of @ajblane is included in this pull request.
Please do not merge it before I fix it.

jserv (changes requested):
Ask @ajblane for final validation.

marktwtn:
I set the thread number of all the testcases to 8.

And the current commit has no problem at all.

wusyong (approved):
The reason for this request is trying to reduce the burden on CI agent. It consumes all cpu when performing the build. It will always be the lower the better IMO. Feel free to choose the proper value you feel comfortable.

ajblane:
Pass test with a broker and a AVX-accelerated remote worker with `make BUILD_REMOTE=1 BIULD_DEBUG=1 check`

jserv:
How about replacing `8` with max CPU cores?

marktwtn:
> How about replacing `8` with max CPU cores?

The original design uses (max CPU cores - 1), which makes our CI agent consume too much CPU resource.

jserv:
> The original design uses (max CPU cores - 1), which makes our CI agent consume too much CPU resource.

Alternatively, we can set its limit to `8` by adopting the available core count.

marktwtn:
> 
> 
> > The original design uses (max CPU cores - 1), which makes our CI agent consume too much CPU resource.
> 
> Alternatively, we can set its limit to `8` by adopting the available core count.

The current design uses the passed in argument as the limit.
In our testcases, it is `8`.
If the available cores are less than the limit, it will use the threads as many as the cores.
Otherwise, the limit will be used.

jserv:
> The current design uses the passed in argument as the limit.
> In our testcases, it is `8`.
> If the available cores are less than the limit, it will use the threads as many as the cores.
> Otherwise, the limit will be used.

The commit messages shall reflect the above.

marktwtn:
The commit message has been updated.

PR #165, opened 2019-06-19
Pre-release v0.4.0

marktwtn:


wusyong (approved):

jserv:
The commits starting with `perf: ` seems to be absent, right?

marktwtn:
> The commits starting with `perf: ` seems to be absent, right?

Yes.
The release robot shows the commits with `feat:` and `fix:` only.

PR #170, opened 2019-07-17
build: Exclude RabbitMQ when BUILD_REMOTE is off

### LINKED ISSUE #169: New updated build won't compile on de10 nano
For some reason I have to build with much older dcurl version on my de10 nano. This is what it states. 
Capture+_2019-07-11-21-49-39

marktwtn:
When the build option BUILD_REMOTE is off, the RabbitMQ should not be
compiled and included in the generated files.
Makefile modification:
- Extract the steps of submodule initialization and update
- Combine and create the submodule related variables

Close #169.

### DIFF
--- a/mk/submodule.mk
+++ b/mk/submodule.mk
@@ -36,9 +36,11 @@ ifeq ($(UNAME_S),darwin)
     LDFLAGS += -L$(OPENSSL_PATH)/lib -lcrypto -lssl
 endif
 
-$(LIBRABBITMQ_PATH)/build/include:
+$(LIBRABBITMQ_PATH)/librabbitmq:
 	git submodule update --init $(LIBRABBITMQ_PATH)
-	mkdir $(LIBRABBITMQ_PATH)/build
+
+$(LIBRABBITMQ_OBJS):
+	mkdir -p $(LIBRABBITMQ_PATH)/build
 ifeq ($(UNAME_S),darwin)
 	# macOS
 	cd $(LIBRABBITMQ_PATH)/build && \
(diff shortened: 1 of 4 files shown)


mk/submodule.mk
```diff
@@ -21,7 +21,7 @@ LIBTUV_BOARD := $(BOARD)
 # PIC (Position-Independent-Code) object file
 LIBTUV_OBJS := $(LIBTUV_PATH)/build/$(LIBTUV_PLATFORM)/release/lib/libtuv.o
 
-$(LIBTUV_PATH)/include:
+$(LIBTUV_PATH)/.git:
```
jserv: I was confused with the transition from `include` to `.git`. Why?

mk/submodule.mk
```diff
@@ -49,7 +51,21 @@ else
          cmake -DCMAKE_INSTALL_PREFIX=. .. && \
          cmake --build . --target install
 endif
-
-$(LIBRABBITMQ_OBJS):
 	cd $(LIBRABBITMQ_PATH)/build && \
          cmake --build .
+
+# Submodules
+SUBS := $(LIBTUV_PATH)/.git
```
jserv: Ditto. We might retrieve packages from archives instead of Git repositories.

mk/submodule.mk
marktwtn: I was thinking that the rule is for detecting when to initialize and update the submodule.
Since not every submodule has the same path to enter the directory `include`, I change it to `.git` to unify it.

I wonder whether there is a better suggestion.

mk/submodule.mk
marktwtn: Got it.

mk/submodule.mk
marktwtn: @jserv should I just revert it?

jserv: > @jserv should I just revert it?

I think so.

mk/submodule.mk
marktwtn: Done.

mk/submodule.mk
marktwtn: Done.

marktwtn:
I forgot to change the `mk/remote.mk` either.
Now it has been done in the forced push commit.

PR #174, opened 2019-07-23
build: Add de10-nano board specific compiler flags

marktwtn:
-mcpu=cortex-a9:
The CPU of de10-nano board is Cortex-A9.
-mfloat-abi=softfp:
The NEON intrinsics are not available with the soft option.
Use softfp instead of soft.
-mfpu=neon:
neon equals to neon-vfpv3.

### DIFF
--- a/mk/board.mk
+++ b/mk/board.mk
@@ -1,0 +1,3 @@
+ifeq ($(BOARD),de10nano)
+    CFLAGS += -mcpu=cortex-a9 -mtune=cortex-a9 -mfloat-abi=hard -mfpu=neon
+endif
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,9 @@ include mk/cpu-features.mk
 # Handle git submodule
 include mk/submodule.mk
 
+# Board specific compiler flags
+include mk/board.mk
+
 # Assign the hardware to CPU if no hardware is specified
 PLATFORMS := $(BUILD_AVX) $(BUILD_SSE) $(BUILD_GENERIC) $(BUILD_GPU) $(BUILD_FPGA_ACCEL)
 ENABLE_PLATFORMS := $(findstring 1,$(PLATFORMS))


mk/board.mk
```diff
@@ -0,0 +1,3 @@
+ifeq ($(BOARD),de10nano)
+    CFLAGS += -mcpu=cortex-a9 -mfloat-abi=softfp -mfpu=neon
```
jserv: Add `-mtune=cortex-a9`
Reference: https://gist.github.com/fm4dd/c663217935dc17f0fc73c9c81b0aa845

jserv: By the way, we can might try `-mneon-for-64bits` option for de10-nano board.

marktwtn: I'll add `-mtune=cortex-a9`.
However, I'm not sure about the necessity of the option `-mneon-for-64bits`.
Based on the GCC manual and the reference, it says: **It is disabled by default due to high cost of moving data from core registers to Neon on 32bit CPU's.**
Out Cortex-A9 CPU is 32-bit.

marktwtn: I have added `-mtune=cortex-a9` and does not include the `-mneon-for-64bits`.

mk/board.mk
```diff
@@ -0,0 +1,3 @@
+ifeq ($(BOARD),de10nano)
+    CFLAGS += -mcpu=cortex-a9 -mtune=cortex-a9 -mfloat-abi=softfp -mfpu=neon
```
jserv: I have no idea why you sticked on `softfp` instead of `hardfp`. Is it limited by prebuilt root file system?

marktwtn: Yes.
If we use `-mfloat-abi=hard`, the compiler says it can not find `gnu/stubs-hard.h` header file and I am sure about that it does not exist on the file system.

jserv: Can you validate with latest Linaro toolchain?

marktwtn: Our **GCC** version is 6.3.0.
The [newest binary release version](https://releases.linaro.org/components/toolchain/binaries/latest-7/arm-linux-gnueabihf/) on the Linaro official website is 7.4.1.
And I do find the file `gnu/stubs-hard.h` in the newest release.
I will download the source code, build and install it.
Or is there any easier way to validate?

jserv: If Debian GNU/Linux is installed on de10-nano, it is always straightforward to validate latest stable packages.

marktwtn: The ARM official website and the Linaro download webpage provide the cross compiler instead of the native compiler.
I'm afraid that I can only validate it with the cross compiler since I found no resource of the native compiler.
And building a native compiler on `DE10 nano` board would be a tedious work.

jserv: For automated builds, we shall always use cross-compilation.

marktwtn: After trying with the hardware floating-point cross compiler, I found out the executables can not be run on `DE10 nano` board since we do not have the corresponding dynamic loader(interpreter).
I think I can mention the `-mfloat-abi=hard` option as a to-do in the comment.

jserv: Can you check this? https://www.digikey.com/eewiki/display/linuxonarm/DE0-Nano-SoC+Kit
I think we shall not support softfp since it is ABI-incompatible with latest Debian GNU/Linux.

marktwtn: I have already followed the steps in the [link](https://www.digikey.com/eewiki/display/linuxonarm/DE0-Nano-SoC+Kit) and even used the newest cross compiler.
In order to validate the `-mfloat-abi=hard`, I think we might need to try another image file rather than the one offered by [LampaLab/iota_fpga](https://github.com/LampaLab/iota_fpga/releases/tag/v0.1)?
We are still lacking of the hardware floating-point related compiler and dynamic loader.

marktwtn: Or maybe try to install those compiler and dynamic loader on the `DE10 nano` board.

jserv: Don't depend on the old image file provided by Lampa Lab since it was indeed out-of-date. We can install Debian-based images on de10-nano and then cross-compile kernel module and dcurl library for usage. hardfp ABI avoids the overhead of moving data between integer and floating point registers but also renders the code incompatible with code built with other -mfloat-abi settings.

jserv: Performance impact caused by softfp: https://github.com/xiph/opus/issues/93

tylerw1369:
You guys can try mine on my repo here https://github.com/tylerw1369/De-10-nano-iota-pow-node.  It has ubuntu armhf and kernel and driver already compiled for armhf.  I had same issues as you guys and had to go armhf to get things to function properly.  The Google download link had the image that is ready to copy to SD card and start up

marktwtn:
> You guys can try mine on my repo here https://github.com/tylerw1369/De-10-nano-iota-pow-node. It has ubuntu armhf and kernel and driver already compiled for armhf. I had same issues as you guys and had to go armhf to get things to function properly. The Google download link had the image that is ready to copy to SD card and start up

@tylerw1369  thanks, it really helps a lot.

mk/board.mk
marktwtn: Based on https://github.com/DLTcollab/dcurl/issues/178, the compiler option `-mfloat-abi=hard` works if we use the armhf based Linux distribution on de10-nano board.

PR #187, opened 2019-08-28
Enable IP address assignment of RabbitMQ broker by user

marktwtn:


src/remote_common.c
```diff
@@ -64,15 +64,16 @@ bool die_on_error(int x, char const *context)
     return true;
 }
 
-bool connect_broker(amqp_connection_state_t *conn)
+bool connect_broker(amqp_connection_state_t *conn, const char *host)
 {
     amqp_socket_t *socket = NULL;
+    const char *hostIP = (host != NULL) ? host : "localhost";
```
jserv: Change from `hostIP` to `hostname`.

src/remote_worker.c
```diff
@@ -6,23 +6,45 @@
...
+    int optIdx;
+    const struct option longOpt[] = {{"broker", required_argument, NULL, 'b'},
+                                     {NULL, 0, NULL, 0}};
 
     amqp_connection_state_t conn;
     amqp_envelope_t envelope;
+    char *hostIP = NULL;
+
+    /* Parse the command line options */
+    /* TODO: Support Mac OS X since getopt_long() is GNU extension */
```
jserv: Change from "Mac OS X" to "macOS" since Apple already rebrands.

src/remote_common.c
jserv: Alternatively, using simple `host` works as well.

src/remote_worker.c
marktwtn: You are right.
I have changed it to `macOS`.

src/remote_common.c
marktwtn: I change the local variable to `host` and the input parameter of `connect_broker()` to `hostName`.
Since the original parameter naming of `amqp_socket_open()` is `host`.

marktwtn:
@jserv the CI has passed all the tests.
If there is nothing to be revised, I think we can merge this pull request?

PR #190, opened 2019-09-06
fix: Release remote interface resource correctly

marktwtn:
Release resource correctly when unexpected things happen such as
the RabbitMQ broker is turned off accidentally.

### DIFF
--- a/src/dcurl.c
+++ b/src/dcurl.c
@@ -126,6 +126,14 @@ int8_t *dcurl_entry(int8_t *trytes, int mwm, int threads)
 
 remote_pow:
     if (!doRemoteContext(&Remote_Context, pow_ctx)) {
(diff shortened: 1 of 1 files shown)


src/dcurl.c
```diff
@@ -126,6 +126,9 @@ int8_t *dcurl_entry(int8_t *trytes, int mwm, int threads)
 
 remote_pow:
     if (!doRemoteContext(&Remote_Context, pow_ctx)) {
+        freeRemoteContext(&Remote_Context, pow_ctx);
+        exitRemoteContext(&Remote_Context);
+        uv_sem_post(&notify_remote);
```
jserv, line 136: Add comments about the termination of remote interface.

marktwtn: Done.

PR #195, opened 2019-09-19
feat: Allocate thread resource based on the physical processor

### LINKED ISSUE #191: Determine the thread pool size based on the physical CPU
The `dcurl` before and after integrating the thread pool uses the **(maximum logical CPU - 1)** as the number of threads for a single PoW.
However, `dcurl` also allows calculating **2** PoW at the same time, which is the number of **threads per core**.
And the number of `logical CPU` is equal to the number of `physical CPU * threads per core`.

Before integrating the thread pool, `dcurl` may create the threads as many as the twice number of the logical CPU.
After integrating the thread pool, `dcurl` set the pool size as **(maximum logical CPU - 1)**.
And it causes the PoW to wait for the threads returning from the previous PoW.

Goal:
- First of all, the thread pool size must be changed to allow **2** PoW being calculated at the same time.
- Second, the thread number should be modified since using **(maximum logical CPU - 1)** number of threads in a single PoW is contradicts to calculating **(threads per core)** number of PoW at the same time.

marktwtn:
The original design has contradiction.
It uses almost all the logical processors for a single PoW,
but allows dcurl to handle 2 PoW at the same time based on the
hyper-threading.

The new function is added in cpu-utils.h to detect the thread number per
physical processor, which also shows whether the hyper-threading is
enabled or not.
Once we know the status of hyper-threading, we can determine the
threadpool size and the thread number for a single PoW.

Close #191.

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -17,6 +17,7 @@ INPUT                  = . \
                          src
 FILE_PATTERNS          = dcurl.c \
                          dcurl.h \
+                         cpu-utils.h \
                          *.md
 USE_MDFILE_AS_MAINPAGE = README.md
 #---------------------------------------------------------------------------
@@ -26,6 +27,10 @@ SOURCE_BROWSER         = YES
 INLINE_SOURCES         = YES
 REFERENCED_BY_RELATION = YES
 REFERENCES_RELATION    = YES
+#---------------------------------------------------------------------------
+# Build related configuration options
+#---------------------------------------------------------------------------
+EXTRACT_STATIC         = YES
 #---------------------------------------------------------------------------
 # Configuration options related to the LaTeX output
 #---------------------------------------------------------------------------
--- a/src/cpu-utils.h
+++ b/src/cpu-utils.h
@@ -15,6 +15,11 @@
 #include <sys/sysinfo.h>
 #endif
 
+/**
+ * @file cpu-utils.h
+ * @brief Utility functions for acquiring CPU information.
+ */
+
 /* On Mac OS X, define our own get_nprocs_conf() */
 #if defined(__APPLE__) || defined(__FreeBSD__)
 #include <sys/sysctl.h>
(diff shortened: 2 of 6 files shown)


marktwtn:
It turns out that the method to acquire processor information differs in different architecture.
I'll figure out another way to implement it on **ARM** and **macOS** system.

src/cpu-utils.h
```diff
@@ -55,3 +68,42 @@ static inline int get_avail_nprocs()
...
+ * @retval 1 Hyperthreading disabled.
+ * @retval 2 Hyperthreading enabled.
+ * @retval -1 Unexpected error.
+ */
+static inline int get_nthds_per_physic_proc()
+{
+    char sib[4], cores[4];
+    int siblings, cpu_cores;
+    FILE *fd;
+    fd = popen("cat /proc/cpuinfo | grep 'siblings' | uniq | awk '{printf $3}'",
```
jserv: It is Linux specific. I would expect both macOS and GNU/Linux can facilitate.

Doxyfile
```diff
@@ -27,6 +28,10 @@ INLINE_SOURCES         = YES
 REFERENCED_BY_RELATION = YES
 REFERENCES_RELATION    = YES
 #---------------------------------------------------------------------------
+# Build related configuration options
+#---------------------------------------------------------------------------
+EXTRACT_STATIC         = YES
```
jserv, line 33: Why would you make such change?

src/pow_sse.c
```diff
@@ -475,7 +476,7 @@ ImplContext PoWSSE_Context = {
     .context = NULL,
     .description = "CPU (Intel SSE)",
     .bitmap = 0,
-    .num_max_thread = 2,
+    .num_max_thread = 0,
```
jserv, line 479: Would you consider to revise its name from `num_max_thread` to `max_thread_num` or `max_thread_per_cpu`?

Doxyfile
marktwtn: To generate the document of static inline function of `cpu-utils.h`.

src/pow_sse.c
marktwtn: Yes.
By the way, the variable name always makes me confused to another similar variable name `num_max_threads` in `pow_avx.h`.
It should be renamed.

jserv: In addition, clarify in the entries in directory `docs/`.

src/cpu-utils.h
marktwtn: I'll change the command to `LC_ALL=C lscpu | grep 'CPU(s)' | awk '{printf $2}'` on GNU/Linux.

marktwtn:
The **Travis CI** does not pass.
It might be caused by the lacking software which should be installed in `.travis.yml`.

PR #198, opened 2019-09-27
Improve CPU information retrieval

marktwtn:


src/pow_c.c
```diff
@@ -339,6 +339,8 @@ static bool PoWC_Context_Initialize(ImplContext *impl_ctx)
 {
     impl_ctx->num_max_thread = get_nthds_per_physic_proc();
     int nproc = get_avail_logic_nprocs() / impl_ctx->num_max_thread;
+    if (nproc == 0)
```
jserv: Can we avoid the duplicated processor information check?

marktwtn: I suggest to change the `get_avail_logic_nprocs()` to `get_avail_physic_nprocs()`.
It would return the `max physical processor - 1` or the least value 1 if no error happened.

jserv: It sounds good for refining `get_avail_physic_nprocs`.

marktwtn: Done.

PR #201, opened 2019-10-03
refactor: Change log function for unification with entangled

### LINKED ISSUE #180: Modify the print out function for better logging
Currently `dcurl` prints out the messages only when it is in **Debug** mode.

For performance measurement and unification with [entangled](https://github.com/iotaledger/entangled), the print out function should be modified.

Goal:
- Enable `dcurl` to print out the logging messages in **Release** mode. And there might need an option to activate the logging system or not.
- Compatible with the logging function in `entangled` since the project [tangle-acclerator](https://github.com/DLTcollab/tangle-accelerator) uses both `entangled` and `dcurl`.

marktwtn:
The original ddprintf() function name is changed to the similar function
log_debug() in entangled.
The implementation is unchanged since the dcurl does not integrate with
the logger project used by entangled.
Currently dcurl has its own implementation.

For unification, the logger_id_t logger_id parameter of log_debug() is added
and assigned with 0.

And there is a new log function log_info() for release mode.

Close #180.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -20,7 +20,13 @@
 
 double diff_in_second(struct timespec t1, struct timespec t2);
 
-static inline void ddprintf(const char *format, ...)
+/* Copy from logger project:
+ * https://bitbucket.org/embear/logger/src/abef6b0a6c991545a3d3fecfbc39d2b0448fb85a/include/logger.h#lines-199*/
+typedef int16_t logger_id_t;
+
+static inline void log_debug(logger_id_t const logger_id,
+                             const char *format,
+                             ...)
 {
 #if defined(ENABLE_DEBUG)
     va_list ap;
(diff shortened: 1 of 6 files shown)


jserv:
Can you check the implementation around `BUILD_STAT` ready to take the revised calls?

marktwtn:
> Can you check the implementation around `BUILD_STAT` ready to take the revised calls?

You mean changing the **printf()** of `BUILD_STAT` enabled testcase to **log_debug()** or **log_info()**?

jserv:
> You mean changing the **printf()** of `BUILD_STAT` enabled testcase to **log_debug()** or **log_info()**?

Exactly.

marktwtn:
@jserv the **printf()** of testcase has been changed to **log_info()** without any problem.

Issue #173, opened 2019-07-21
Integrate static code analysis into CI


marktwtn:
Currently we have dynamic code analysis such as **GCC Sanitizers**.
To have a better code quality, we should use the static code analysis tool such as **Cppcheck** and **Infer**.

marktwtn:
As I was trying to eliminate the errors detected by **Cppcheck**, the error messages showed that the functions in **compat-ccurl.c** were not called by any other function.

Since dcurl had integrated with **GCC Sanitizer**, we removed the testing python code to avoid error when running with **Sanitizer**.
It also removed the code of testing **compat-ccurl.c**.
And the build option `BUILD_COMPAT` was marked as **deprecated** in the document.

Shall we be going to remove the whole **compat-ccurl.c**?
Is it designed to be compatible with the **ccurl** in `IOTA wallet`?
And is supporting `IOTA wallet` and **ccurl** necessary?

jserv:
You can exclude `compat-ccurl.c` for cppcheck listing. It would be still practical to ensure CCurl compatibility though.

marktwtn:
I recommend to download the newest **CppCheck** release source code from the [official website](http://cppcheck.sourceforge.net/),
since `$ apt install cppcheck` command would install outdated version and it has some bugs.

The easiest steps to install **CppCheck** from the source code are:
```
$ make FILESDIR=/usr/bin
$ sudo make install FILESDIR=/usr/bin
```

marktwtn:
On Linux, **Infer** does not offer quick installation with the command `$ apt install`.
Please follow the [reference](https://fbinfer.com/docs/getting-started.html) to install it from the binary release.

marktwtn:
I pushed a branch for infer: https://github.com/DLTcollab/dcurl/tree/infer

It works fine on my desktop but fails on the CI environment.
The error message from Makefile says `javah is not available. Please check JDK installation`.
However, it only happens when we do the static code analysis.
If we just build dcurl with `BUILD_JNI` option enabled and do not do the static code analysis, everything goes well.

jserv:
> It works fine on my desktop but fails on the CI environment.
> The error message from Makefile says `javah is not available. Please check JDK installation`.

Does it imply the Insufficient detections for JNI specific tools?

marktwtn:
> Does it imply the Insufficient detections for JNI specific tools?

It implies **Infer** is an interesting tool.
[...]

PR #202, opened 2019-10-16
Cppcheck integration

marktwtn:
The detected and fixed errors list:
- Assign a value that is never used
- Reassign a value before the old one has been used
- The scope of the variable can be reduced
- The behaviour of using void pointers in calculations is undefined
- Uninitialized variable
- Shadow variables

The file cppcheck_suppress suppresses the errors which should not be
treated as errors.
The reason is written in the file.

The sanitizer.mk is renamed as dynamic-analysis.mk.
And there is another file static-analysis.mk for static code analysis.

Related to #173.


docs/build-n-test.md
```diff
@@ -132,6 +132,12 @@ Success.
         [ Verified ] 
 ```
 
+## Static Code Analysis
+### CppCheck
+```
+$ make cppcheck
```
jserv: We shall not specify the target which is mapped to exact tool names. Instead, the build target might be `static-analysis`.

marktwtn: My original design is to have two target, one is for `CppCheck` and one is for `Infer`.
I think I will modify the target to `static-analysis`, and it would use `CppCheck` and `Infer` to do the static code analysis.

jserv: Let's blur the targets such as `cppcheck` and `infer` because they might be really functioned under certain orders or conditions.

marktwtn: Done.

PR #213, opened 2020-01-03
Naming convention

### LINKED ISSUE #207: Unify the naming convention with snake case
The variable and function naming is messy.
One might use `camel case`, and another might use `snake case` and sometimes the other might use `upper camel case`.

In the future, the `dcurl` may be integrated into [entangled](https://github.com/iotaledger/entangled).
We should follow the same naming convention, which is `snake case` in my observation.

marktwtn:
style: Unify the naming convention with snake case

Besides the source code unification, the related document is added
for the naming convention details.
Close #207.

style: Revise source code for coding style


docs/naming-convention.md
```diff
@@ -0,0 +1,27 @@
...
+  ```
+  typedef struct pwrok_s pwork_t;
+  struct pwork_s {
+      ......
+  }
+  ```
+
+- Macro: Use capital letters
+  ```
+  #define MIN_TRYTE_VALUE -13
```
jserv, line 24: Should be `(-13)` rather than `-13`.

marktwtn: I'll fix it in another pull request for adding the bracket to macros.

PR #222, opened 2020-03-05
feat: Change dcurl_init API to enable initialization configuration

### LINKED ISSUE #220: Assign RabbitMQ broker IP address at the dcurl publisher side
When the build option `BUILD_REMOTE` is enabled, the **dcurl** can be treated as two parts.
The library for local machine and the executable for remote machine.
That is, the publisher and the consumer.

In https://github.com/DLTcollab/dcurl/issues/185, the consumer can assign the **RabbitMQ broker** IP.
However, the publisher can not.

The default IP value on the publisher side is `localhost`.
It means that the **dcurl** publisher and the **RabbitMQ broker** should be on the same machine only, which is not flexible.

To allow the **dcurl** publisher to assign the **RabbitMQ broker** IP address, there are two methods:
- Assign the IP address at the compilation stage.
- Assign the IP address when the **dcurl** is initializing.

marktwtn:
The dcurl_init API can pass a pointer to enable initialization
configuration.
Currently it only sets the hostname of the broker.
If the passed pointer is NULL, the pre-defined default value would be used.

The API document description and the test cases are modified as well.

Close #220.

### DIFF
--- a/jni/iri_pearldiver_exlib.c
+++ b/jni/iri_pearldiver_exlib.c
@@ -13,11 +13,32 @@
 #include "../src/trinary.h"
 
 JNIEXPORT jboolean JNICALL
-Java_com_iota_iri_crypto_PearlDiver_exlibInit(JNIEnv *env, jclass clazz)
+Java_com_iota_iri_crypto_PearlDiver_exlibInit(JNIEnv *env,
+                                              jclass clazz,
+                                              jbyteArray broker_host,
+                                              jint broker_host_len)
 {
-    if (!dcurl_init())
-        return JNI_FALSE;
-    return JNI_TRUE;
+    jboolean ret = JNI_TRUE;
+
+    /* Get the Byte array from Java byte Array */
+    jbyte *host = (*env)->GetByteArrayElements(env, broker_host, NULL);
+
+    dcurl_config config;
+    config.broker_host = malloc(broker_host_len + 1);
+    if (!config.broker_host) {
+        ret = JNI_FALSE;
+        goto fail;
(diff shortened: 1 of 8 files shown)


jni/iri_pearldiver_exlib.c
```diff
@@ -13,11 +13,33 @@
...
+                                              jclass clazz,
+                                              jbyteArray broker_host,
+                                              jint broker_host_len)
 {
-    if (!dcurl_init())
-        return JNI_FALSE;
-    return JNI_TRUE;
+    jboolean ret = JNI_TRUE;
+
+    /*********** Get the Byte array from Java byte Array *************/
```
jserv: Avoid using comment redundant comment style. Use plain `/* The content */` instead.

src/dcurl.h
```diff
@@ -23,16 +23,32 @@
...
 
+/*! The default value of the broker hostname */
+#define DEFAULT_BROKER_HOST "localhost"
+
+/**
+ * A structure representing the configuration of the initialization.
+ */
+typedef struct {
+    char *broker_host; /**< The broker hostname used in the remote mode */
+} init_config;
```
jserv: Don't name it `init_config`. Instead, use `dcurl_config`.

marktwtn: Done.

jni/iri_pearldiver_exlib.c
marktwtn: I also refactor the other comments in `jni/iri_peraldiver_exlib.c`.

PR #225, opened 2020-03-08
docs: Add rule of document creation and deletion

marktwtn:


Makefile
```diff
@@ -41,6 +41,9 @@ include mk/submodule.mk
 # Board specific compiler flags
 include mk/board.mk
 
+# Generate document
+include mk/doc.mk
```
jserv: Rename it to `docgen` or `doxygen` to avoid implications.

marktwtn: Rename it as `docgen.mk`.

